PR #13, opened 2016-10-13
Made generated class names shorter

scott113341:
This will fail CI for sure because I didn't change all of the tests.  Only tests in `styletron-core` are changed.  I will change the rest if this looks good to you!

gzipped size comparison incoming...

### DIFF
--- a/packages/styletron-core/src/class-name.js
+++ b/packages/styletron-core/src/class-name.js
@@ -1,0 +1,24 @@
+// numbers are NOT allowed as first characters of CSS class names
+const firstChars = Array.from(Array(26), (a, i) => (i + 10).toString(36));
+
+/**
+ * Generates a short, valid CSS class name
+ * @param int {int}  A number
+ * @returns {string} CSS class name
+ */
+function className(int) {
+  if (int < 26) {
+    return firstChars[int];
+  } else {
+    const i = int - 26; //
+    const namespaceN = Math.floor(i / 36);
+    const namespaceIndex = namespaceN % 26;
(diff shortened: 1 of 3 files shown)


scott113341:
With this setup:

```
const CSSOM = require('cssom');
const fs = require('fs');
const gzipSize = require('gzip-size');

const StyletronServer = require('./packages/styletron-server/lib/styletron-server.js');

const css = fs.readFileSync('./packages/benchmarks/fixtures/css/uber.css').toString();
const parsedCss = CSSOM.parse(css);

const server = new StyletronServer();
let totalRules = 0;

parsedCss.cssRules.forEach(rule => {
  if (rule.style) {
    for (let i = 0; i < rule.style.length; i++) {
      const prop = rule.style[i];
      const val = rule.style[prop];
      server.injectDeclaration({ prop, val });
      totalRules++;
    }
  }
});

const html = server.getStylesheetsHtml();
const size = gzipSize.sync(html);

console.log('total rules:', totalRules);
console.log('unique rules:', server.uniqueCount);
console.log('original length:', css.length);
console.log('styletron length:', html.length);
console.log('gzipped length:', size);
```

With old method (c0, c1, ...):
```
total rules: 2460
unique rules: 1016
original length: 282167
styletron length: 31847
gzipped length: 7540
```

With new method (a, b, c):
```
total rules: 2460
unique rules: 1016
original length: 282167
styletron length: 30895
gzipped length: 7554
```

So my method is actually worse once gzipped, despite being 3% smaller in terms of raw HTML output by `styletron-server`.  I'm not sure why though...

PR #20, opened 2016-12-06
Add media attribute to client mediaSheets

### LINKED ISSUE #19: Missing client-side media attribute
Client-side media queries don't work because the mediaSheets lack a matching media attribute [https://github.com/rtsao/styletron/blob/master/packages/styletron-client/src/styletron-client.js#L81-L83](https://github.com/rtsao/styletron/blob/master/packages/styletron-client/src/styletron-client.js#L81-L83)

PR on its way 😄

stevenbenisek:
Fixes #19


rtsao:
Awesome! thank you for the fix

PR #25, opened 2016-12-07
StyletronCore constructor defaults and docs

scott113341:
I did not update the docs for `StyletronServer` because its method signature is identical to that of `StyletronCore` (and it is already documented as extending `StyletronCore`).


scott113341:
For reference, this is what the resulting docs look like: https://announcer-ape-14688.netlify.com

PR #29, opened 2016-12-08
Added cname cli to create CNAME file from package data

cjpatoilo:
Hi @rtsao 

This lib reads the package.json files to get the homepage value, then create the CNAME file with that value. This lib covers any problems with environments because it was written in Node.js. For more information click [read](https://github.com/cjpatoilo/cname).


rtsao:
The CNAME file only needs to be generated on Travis, so I don't think there's any cross-platform considerations here. I'm not sure it'd be worth adding an extra dependency so I'm going to close this for now.

PR #30, opened 2016-12-08
Updated badge by changing dependencies to devdependencies

cjpatoilo:
Hi @rtsao 

I submitted this pull request to fix the dependencies badge. There are no `dependencies`, just `devDependencies`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,7 +132,7 @@ function render() {
   );
 
   const stylesForHead = styletron.getStylesheetsHtml();
-  
+
   return `<html><head>${stylesForHead}</head><body>${appMarkup}</body></html>`;
 }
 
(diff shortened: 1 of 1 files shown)


rtsao:
Yeah, the dependency badge situation is interesting because Styletron is a multi-package monorepo.

I don't imagine any styletron packages having any (non-styletron monorepo) dependencies ever because the focus is so narrow, and Lerna automatically keeps these up to date, so maybe these are redundant? But I suppose it would probably be a good idea to have dependency badges in each package README anyways, just to reinforce that they are kept up-to-date.

I'm not quite sure what I think the right course of action would be for the main README. Other monorepos have tables of badges, [1] and [2] are good examples, but for something as focused as Styletron I'd be concerned about it bringing too much noise to the README. My gut feeling is that it would need to be to the point where there's a better canonical documentation/intro website, and the GitHub README is more contributor-focused.

[1] https://github.com/babel/babel#core-packages
[2] https://github.com/cyclejs/cyclejs#packages

rtsao:
As discussed above, I think this is a somewhat large task due to our monorepo setup. Closing for now.

PR #32, opened 2016-12-08
Added github templates

cjpatoilo:
Hi @rtsao

It's hard to solve a problem when important details are missing. Now project maintainers can add templates for Issues and Pull Requests to projects, helping contributors add the right details at the start of a thread. For more information click [here](https://github.com/blog/2111-issue-and-pull-request-templates).

### DIFF
--- a/.github/contributing.md
+++ b/.github/contributing.md
@@ -1,0 +1,49 @@
+# Contributing
+
+Contribute to Styletron and help us make this even better! Start reading this document to see it is not difficult as you might have imagined.
+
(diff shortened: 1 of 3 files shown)


rtsao:
You bring up good points, but I think it's worth having more tailored templates specifically for Styletron as these seem a bit wordy. There hasn't been a flood of issues, so I think everything has been manageable so far. Definitely worth revisiting in the future though if more issues keep cropping up. Closing for now.

PR #37, opened 2016-12-09
Fixed #36

### LINKED ISSUE #36: Components returned from "styled" have no backing instance, preventing refs from being set
For example,

```
class Parent extends Component {
  render() {
    return (
      <div>
        // ChildComponent is a styletron component.
        // this.childComponent is undefined
        <ChildComponent ref={(c) => { this.childComponent =  c; }}
      </div>
    );
  }
}
```

sammkj:
Usage:

```
class Parent extends Component {
  render() {
    return (
      <div>
        // innerRef not ref
        <ChildComponent innerRef={(c) => { this.childComponent =  c; }}
      </div>
    );
  }
}
```

### DIFF
--- a/packages/styletron-react/src/styled.js
+++ b/packages/styletron-react/src/styled.js
@@ -63,26 +63,57 @@ function styled(base, styleArg) {
 
 function createStyledElementComponent(tagName, stylesArray) {
 
-  const StyledElement = (props, context) => {
-    
-    const resolvedStyle = {};
-    StyledElement[STYLES_KEY].forEach(style => {
-      if (typeof style === 'function') {
-        assign(resolvedStyle, style(props, context));
-      } else if (typeof style === 'object') {
-        assign(resolvedStyle, style);
+  class StyledElement extends React.Component {
+    constructor(props) {
+      super(props);
+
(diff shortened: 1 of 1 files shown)


packages/styletron-react/src/styled.js
```diff
@@ -63,26 +63,57 @@ function styled(base, styleArg) {
...
+        elementProps,
+      });
+    }
 
-    const elementProps = omitInvalidProps(props);
-    elementProps.className = props.className
-      ? `${props.className} ${styletronClassName}`
-      : styletronClassName;
+    render() {
+      const {elementProps} = this.state;
```
rtsao: Out of curiosity, is there an advantage to using lifecycle methods/state rather than keeping everything in render?

sammkj: With lifecycle, we could do some optimization if needed. I was thinking of implementing something to compare the `resolvedStyle` in `componentWillReceiveProps`.

rtsao: Gotcha, that's a good idea to look into.

rtsao:
Thanks for the PR! 😄

PR #41, opened 2016-12-11
Fix for chrome CSSOM bug

tiagojsalmeida:
Using the `appendChild` instead of `insertRule` fixes the chrome bug at least.
Were you using `insertRule` for performance reasons @rtsao ?


Cavitt:
`insertRule` should be used in production, `appendChild` in development.

rtsao:
I'd ideally like to avoid differences in prod vs. dev as much as possible (including perf optimzations). I'm currently leaning the route of https://github.com/rtsao/styletron/pull/111 to solve debugging problems instead.

PR #52, opened 2016-12-18
Add TypeScript declaration files

faddee:
This automatically adds TypeScript support for `styletron-core`, `styletron-server`, `styletron-client` and `styletron-react`.


rtsao:
Awesome, thanks!

PR #54, opened 2016-12-21
Add support for cssFloat

vikfroberg:
When using `postcss-js` it transforms `float` to `cssFloat`, currently styletron doesn't handle that.

### DIFF
--- a/packages/styletron-utils/src/hyphenate-style-name.js
+++ b/packages/styletron-utils/src/hyphenate-style-name.js
@@ -5,6 +5,7 @@ module.exports = hyphenateStyleName;
 
 function hyphenateStyleName(prop) {
   return prop
+    .replace('cssFloat', 'float')
     .replace(uppercasePattern, '-$&')
     .toLowerCase()
     .replace(msPattern, '-ms-');


packages/styletron-utils/src/hyphenate-style-name.js
```diff
@@ -5,6 +5,7 @@ module.exports = hyphenateStyleName;
 
 function hyphenateStyleName(prop) {
   return prop
+    .replace('cssFloat', 'float')
```
rtsao: I think a simple conditional, e.g.

```js
if (prop === 'cssFloat') {
  return 'float';
}
```

Would probably be faster.

rtsao:
Interesting! I did not know about `cssFloat` before. It seems both Chrome and React both support `float` style properties, not just `cssFloat`.

I'm somewhat hesitant to add this functionality due to the potential performance overhead for a rare use-case, since it seems to me most of the time developers will use `float` in their object literals and hyphenate-style-name is a very hot code path.

I actually somewhat take issue with this functionality in postcss-js, since it's designed to work with React and CSS-in-JS libraries, not bona fide inline style property manipulation. Since it supports media queries and pseudo selectors (like most CSS-in-JS libs) compatibility with inline style property getter/setters gets thrown out the window (which is the source of the`cssText` quirk). I think ideally postcss-js should change it's behavior.

vikfroberg:
> I did not know about cssFloat before

Me neither to be honest! 

> I actually somewhat take issue with this behavior in postcss-js, since it's designed to work with React and CSS-in-JS libraries, not bona fide inline style property manipulation

Yeah, maybe. 

React supports `cssFloat` though. Here are some PR's for reference.
https://github.com/facebook/react/pull/6133
https://github.com/postcss/postcss-js/issues/11

I'll create an issue for `postcss-js` repo as well.

vikfroberg:
It seems that `postcss-js` isn't interested in changing it's behaviour. Since you implemented a cache, do you stil think it's to expensive to add an if statement for `cssFloat`?

vikfroberg:
Middlewares will solve this.

PR #59, opened 2016-12-25
Make stylesheets optional in StyletronClient constructor

### LINKED ISSUE #34: Make stylesheet(s) optional in StyletronClient constructor
Per https://github.com/rtsao/styletron/issues/22, as Styletron no longer uses just a single stylesheet, it doesn't make sense for the user to pass it in for client-side only use-cases. Instead, if no server-rendered stylesheet is provided as an argument, one should be created and appended to the document `<head>`

Gandem:
Hey !

I guess this fixes #34 

This is one of my first contributions to OSS, so if my PR is obviously wrong, I'll be glad to fix it 😃


rtsao:
Awesome, thank you so much!

PR #60, opened 2016-12-26
Add the render target for the client-side rendering example

sergiodxa:
The client-side rendering example with React lack the render target node so it can't be copy-pasted to test it.


rtsao:
Thank you!

PR #64, opened 2017-01-03
Pass `className` prop to components

faddee:
This makes the `styled` function accept any component type. It behaves like before if you pass a tag name string or a styled component as first argument. But if you pass another component, it simply passes all props and adds the generated class names to the `className` prop and assumes that the component handles that manually.


rtsao:
LGTM, thank you! 😄

PR #65, opened 2017-01-04
Typescript changes

faddee:
Changed TypeScript declarations to support any component type as argument for the `styled` function #64. What also includes in this PR is the optional argument for the `StyletronClient` constructor #59 and a hack to enable autocompletion for CSS properties as object.


faddee:
Also added a generic `styled<TElement, TProps>` with `base` as string for the ability to manually type when you have e.g. `base` as a variable.

```tsx
type PropType = {
  bold: boolean;
}

function factory<TElement extends HTMLElement>(base: string, color: string) {
  return styled<TElement, PropType>(base, props => ({
    color,
    fontWeight: props.bold ? 'bold' : 'normal',
  }));
}

const RedDiv = factory<HTMLDivElement>('div', 'red');
const GreenSpan = factory<HTMLSpanElement>('span', 'green');
```

rtsao:
LGTM, thanks!

PR #67, opened 2017-01-09
Proposal: Merge `styled` style arguments

faddee:
This proposal changes the `styled` function to be able to receive multiple style arguments and merges these for you. The background for this is that you often need to make conditional styles and not only toggle between values when you have more complex styling.

```js
styled('div', ({center, children}) => {
  return Object.assign(
    {
      // Default styling
      color: 'red',
      background: 'black',
    },
    React.Children.count(children) > 1 && {
      display: 'flex',
    },
    center && {
      [React.Children.count(children) > 1 ? 'alignContent' : 'textAlign']: 'center',
    }
  );
});
```

The reason I send you this proposal is because we need to assign the styling objects manually like in the example above which causes some unnecessary overhead, since `styled` already do this, and makes the code less pretty. ;-)

```js
styled('div',
  {
    // Default styling
    color: 'red',
    background: 'black',
  },
  ({children}) => React.Children.count(children) > 1 && {
    display: 'flex',
  },
  ({center, children}) => center && {
    [React.Children.count(children) > 1 ? 'alignContent' : 'textAlign']: 'center',
  }
);
```

The change is minor, but I'm aware that this could cause some trouble in the future if you for some reason would like to have another argument for `styled`. But I hope you agree that this would be a nice feature to offer and if this issue regarding all arguments being occupied occurs, that it would be solvable with feature functions like:

```js
import {styled, aNiceFeature} from 'styletron-react';
styled(aNiceFeature('div'), {color: 'red'});
```

Edit: Another reason for us who are using TypeScript, with `Object.assign` you need to define type manually to get autocompletion and typechecking. This way you get it for free.


rtsao:
Interesting proposal! I've been pretty extensively using object spread syntax for conditional styles:
```js
styled('div', ({center, children}) => ({
  // Default styling
  color: 'red',
  background: 'black',
  ...(React.Children.count(children) > 1 && {
    display: 'flex',
  }),
  ...(center && {
    [React.Children.count(children) > 1 ? 'alignContent' : 'textAlign']: 'center',
  })
}));
```
I think it's not *too* horrible in most cases. What if an array was accepted as a second argument? Then more arguments could be provided down the road, but still have the automatic merging functionality like you describe.

faddee:
The spread operator makes it look pretty good. I kind of like the idea that the operator acts like an indicator that a condition is happening. It gives you a great overview. However, TypeScript doesn't get that the spread object is of type `React.CSSProperties` anyway.

I though of accepting an array as second argument too, but I believe it would be a prettier implementation detail to rest the arguments as style arguments rather than having to wrap it as an array. But if you prefer to be able to have a third argument, I would be fine with an array.

faddee:
Which one is more preferable? One of them or both?

```js
styled('div', [
  {...},
  props => ({...}),
  props => ({...}),
]);
```
... or...

```js
styled('div', props => [
  {...},
  {...},
  {...},
]);
```

faddee:
Closing this in favor for spread operator.

PR #74, opened 2017-01-21
Remove prop whitelist

apalm:
Looking through the `styletron-react` source, I noticed the `omitInvalidProps` function. I assume this is for the [unknown prop warning](https://facebook.github.io/react/warnings/unknown-prop.html).

This PR removes the whitelist, as using whitelists is [not recommended](https://twitter.com/dan_abramov/status/822526388557250560).


rtsao:
The thing with the "styled components" abstraction is your styled component is *just* an element with styles.

```jsx
const FancyDiv = styled('div', props => ({color: props.someProp}));

<FancyDiv someProp="red" id="some-valid-div-attr" data-arbitrary="valid-attr"/>
```

Our style function takes props, which means we can build abstraction. However, in this case, `someProp` is not a valid HTML attribute, which will trigger this warning. So we only want to pass through valid HTML attributes to the underlying ReactElement.

So just like styled-components, we need [this whitelist](https://github.com/styled-components/styled-components/blob/master/src/utils/validAttr.js).

The whitelist could be avoided by doing something like this:
```jsx
const FancyDiv = styled('div', props => ({color: props.someProp}));

<FancyDiv styleProps={{someProp: "red"}} id="some-valid-div-attr" data-arbitrary="valid-attr"/>
```

But I from a developer experience perspective this seems much worse. I think the nice thing about styled components is that props are just props.

apalm:
Ah, I knew I was missing something 😅 Thanks for the explanation.

I agree that the DX is worse in your second example above. Feel free to close this PR if you don't want to take it.

rtsao:
No problem!

It's possible to use static analysis to automatically generate a prop blacklist (instead of including a whitelist). I've built a proof of concept that is fairly robust, but if there's no static access then false negatives will occur. In theory, the size of the accumulated blacklist would be much smaller than the whitelist (in most conceivable real-world cases).

The major downside to this approach is that a Babel plugin would be required. That said, it's something I'm still interested in pursuing at some point.

Issue #73, opened 2017-01-20
Default dynamic props


vikfroberg:
StyledComponents have recently been discussing dynamic default props: https://github.com/styled-components/styled-components/pull/365

It would be pretty useful for styletron as well I think, maybe a third argument?

vikfroberg:
I feel like the most important use case here is the ability to generate style attributes from props for high performance styles. https://github.com/styled-components/styled-components/issues/134#issuecomment-257301370

Also https://github.com/rtsao/styletron/tree/master/packages/babel-plugin-styletron this might not be a good idea because of that?

faddee:
I think it would be better to have a prop function as argument to `styled` that invokes on each render and get generated props argument.

```js
styled(
  'div',
  ({warn}) => ({color: warn ? 'red' : 'green'}),
  ({warn, className, anotherProp, ...restProps}) => ({
    ...restProps,
    classes: className,
    anotherProp: anotherProp || 'default-value',
  })
)
```

In that way you have the ability to filter out props that you don't wan't to be passed (`warn` in this case), map `className` prop to another prop and add default values.

faddee:
In the high performance issue you had as a exaple it would look something like this:

```js
styled(
  'div',
  { /* Lots of style */ },
  ({hue, luminosity, ..restProps}) => ({
    ...restProps,
    style: {
      background: `hsl(
        ${props.hue}, 
        ${50 + props.luminosity / 2}%, 
        ${props => props.luminosity}%
      )`,
    },
  })
)
```

vikfroberg:
I'm not sure about forcing the user to forward all props. Do you guys think omitting props that are not valid, is a bad practise?

faddee:
You cannot automatically omit props when you pass a component as first argument. And sometimes that's troubling because that component may spread props to an element and you end up with "invalid props" warning if you don't filter it out manually with yet another component. And we usually use the prop `as` to an element to toggle different appearances of the element. But `as` is not automatically omitted by some reason and ends up in the DOM. So I like the idea of having a function the gives you the opportunity to manipulate props before they get passed to my base element or component.

faddee:
The function would of course be optional. So if you don't have a third argument, props will be passed as they are. But if you use it, you would have to spread the rested props or assign them, but I believe that detail is fine since you do it all the time on other places like in reducers.

rtsao:
Closing for inactivity. Feel free to re-open if this is still an issue. Thanks!

PR #79, opened 2017-01-25
Improve TypeScript declaration

faddee:
After a couple of weeks of testing I found some improvements that had to be made:

 - **SVG elements receives SVG-attributes as props**
   Fixed SVG-elements like `svg` and `g`  where the returned component from `styled` received HTML-attributes instead of SVG-attributes as props and therefor didn't accept props like `viewport` or `cx`.
 - **Pass instance of base component as argument through `innerRef`**
   When the base is a class component, the type of the `innerRef` argument is now the instance of the real base class instead of a generic component.
   ```ts
   class Foo extends React.Component<{}, void> {
     someProperty = "bar";
     render() {
       return <div />
     }
   }
   
   const Bar = styled(Foo, {});

   <Bar innerRef={foo => foo.someProperty === 'bar'} />
   ``` 
 - **Prop type for base component and style function should match**
   When you have a component as base, prop type for the component and the style function should match because props aren't emitted. This would cause a "invalid prop"-warning of `warn`:
   ```ts
   const Shared = (props: React.HTMLAttributes<HTMLDivElement>) => <div {...props} />;
   
   const Foo = styled(Shared, ({warn = false}) => ({
     background: warn ? 'red' : 'green',
   }));
   
   <Foo warn />
   ```
   Now here's a change so TypeScript doesn't allow you to do this. With no capabilities to filter out `warn` natively, you have to do something like this:
   ```ts
   const Shared = (props: React.HTMLAttributes<HTMLDivElement>) => <div {...props} />;

   const Foo = styled<React.HTMLAttributes<HTMLDivElement> & {warn: boolean}>(
     ({warn, ...rest}) => <Shared {...rest} />,
     ({warn = false}) => ({
       background: warn ? 'red' : 'green',
     })
   );
   
   <Foo warn />
   ```
   It's not pretty, but TypeScript should tell you the truth so you don't run into React-warning surprises. Hopefully #73 will result into a native way of doing this with Styletron in the future. :-)

I would appreciate if you could publish a patch release after this @rtsao ❤

### DIFF
--- a/packages/styletron-react/index.d.ts
+++ b/packages/styletron-react/index.d.ts
@@ -1,209 +1,247 @@
-import {Component, ComponentClass, StatelessComponent, DOMElement, ComponentElement, SFCElement, ComponentState, CSSProperties, HTMLAttributes, Ref} from "react";
+import * as React from "react";
 import StyletronServer from "styletron-server";
 import StyletronClient from "styletron-client";
 
 declare namespace StyletronReact {
-  export class StyletronProvider extends Component<{styletron: StyletronServer | StyletronClient}, void> {}
+  export class StyletronProvider extends React.Component<{styletron: StyletronServer | StyletronClient}, void> {}
 
   type InnerRef<TInstance> = {
-    innerRef?: Ref<TInstance>;
+    innerRef?: (instance: TInstance) => any;
   };
 
-  interface StyledElement<TProps, TElement extends Element> extends ComponentClass<TProps & HTMLAttributes<TElement> & InnerRef<TElement>> {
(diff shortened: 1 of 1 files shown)


faddee:
For your information. Another known problem is that TypeScript doesn't warn if the type property `className` isn't `className?: string` the base component (not element). I haven't figured out any way to solve this yet. TypeScript gets confused during tests if I do `TProps & {className: string}` and falls back to another type of `styled` and warns about a lot of other stuff which wouldn't help you as a developer.

rtsao:
Nice work, thank you! 😄

PR #80, opened 2017-01-25
Don't pass `innerRef` prop

faddee:
As `ref` is omitted by React, the same thing should happen with `innerRef`. This example would cause "invalid prop"-warning since invalid props are not omitted when passing a component as argument to `styled`.

```ts
class Shared extends React.Component {
  render() {
    return <div {...this.props} />
  }
}

const Foo = styled(Shared, {
  color: 'pink',
});

<Foo innerRef={shared => shared} />
```

This is not an issue when the argument to `styled` is a string, but this PR makes sure `innerRef` is omitted no matter what.


rtsao:
Ah, nice catch. Thanks!

PR #90, opened 2017-02-17
Added bindings for Inferno

wagerfield:
It's finally here...Styletron bindings for [Inferno](https://infernojs.org/) 🎉 

Par a few minor differences, this package behaves in exactly the same way as `styletron-react`. 

The notable differences are as follows:

1. `StyletronProvider` is just called `Provider`—the majority of other bindings out there for Redux, MobX etc. all use the non-prefixed `Provider` name, so I thought we should follow convention. If the user is using a number of providers, they can simply name them accordingly when they `import/require` them.
2. The `styled` function params are called `name` and `styles` as opposed to `base` and `styleArg`. `name` is the param name used in Inferno's `createElement` function and `styles`...well just because I don't think `Arg` is needed :)
3. I omitted the `isValidAttr` parser because Inferno handles this internally. Unless I'm missing something, I don't think it's necessary.
4. I made a few optimisations to the `styled` implementation (such as using for loops over lambdas and using a functional component rather than a class component to return the styled component).
5. I also stored the `name` and `stylesArray` on a single `__STYLETRON` object instead of 2 separate keys (`__STYLETRON_STYLES` and `__STYLETRON_TAG`) because I thought it was a little neater.
6. I created some generic `utils` like `assign` and a few type checkers like `isString`, `isObject` etc. These are obviously very generic, so if you're doing these checks anywhere else, perhaps we could hoist them into `styletron-utils` for other `styletron` packages to use and reduce the footprint?

In addition to the above, I have also written a pretty comprehensive `README` to compliment the JavaDocs that I have added in `provider.js` and `styled.js`.

Perhaps the changes above could be rolled into the React version? Happy to do that if you're in agreement.

This PR resolves https://github.com/rtsao/styletron/issues/50

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ npm-debug.log.*
 htmldocs/
 lerna-debug.log
 lib/
+.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,3 @@ npm-debug.log.*
 htmldocs/
 lerna-debug.log
 lib/
-.DS_Store
--- a/packages/styletron-inferno/src/styled.js
+++ b/packages/styletron-inferno/src/styled.js
@@ -3,8 +3,8 @@ const {injectStylePrefixed} = require('styletron-utils');
 const {
   assign,
   isNil,
-  isString,
   isObject,
+  isString,
   isFunction
 } = require('./utils');
 
@@ -87,7 +87,7 @@ function createStyledComponent(name, stylesArray) {
 
 function resolveStyles(stylesArray, props, context) {
   const resolvedStyles = {};
-  for (let styles, i = 0, l = stylesArray.length; i < l; i++) {
+  for (let i = 0, l = stylesArray.length, styles; i < l; i++) {
     styles = stylesArray[i];
     if (!isNil(styles)) {
       if (isFunction(styles)) {
--- a/packages/styletron-inferno/src/styled.js
+++ b/packages/styletron-inferno/src/styled.js
@@ -1,0 +1,124 @@
+const createElement = require('inferno-create-element');
+const {injectStylePrefixed} = require('styletron-utils');
+const {
+  assign,
+  isNil,
+  isString,
+  isObject,
+  isFunction
+} = require('./utils');
+
+const STYLETRON_KEY = '__STYLETRON';
+
+module.exports = styled;
+
+/**
+ * Helper function to create styled components
+ * @packagename styletron-inferno
+ * @param  {string|function} name   Tag name or component function/class
+ * @param  {function|object} styles Style object or function that returns a style object
+ * @return {function}               Styled component
+ *
+ * @example
+ * import { styled } from 'styletron-inferno';
+ *
+ * const Panel = styled('div', {
+ *   backgroundColor: 'lightblue',
+ *   fontSize: '12px'
+ * });
+ *
+ * <Panel>Hello World</Panel>
+ *
+ * @example
+ * import { styled } from 'styletron-inferno';
+ *
+ * const Panel = styled('div', (props) => ({
+ *   backgroundColor: props.alert ? 'orange' : 'lightblue',
+ *   fontSize: '12px'
+ * }));
+ *
+ * <Panel alert>Danger!</Panel>
+ *
+ * @example
+ * import { styled } from 'styletron-inferno';
+ *
+ * const DeluxePanel = styled(Panel, (props) => ({
+ *   backgroundColor: props.alert ? 'red' : 'lime',
+ *   boxShadow: '3px 3px 3px darkgray',
+ *   color: 'white'
+ * }));
+ *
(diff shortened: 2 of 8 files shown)


rtsao:
Awesome work! Overview of changes sounds good, I'll review this shortly 👍

packages/styletron-inferno/src/styled.js
```diff
@@ -0,0 +1,124 @@
...
+        assign(resolvedStyles, styles);
+      }
+    }
+  }
+  return resolvedStyles;
+}
+
+const SPACE_REGEX = /\s+/;
+
+function mergeClassNames(classNameA, classNameB) {
```
rtsao: I think the addition of de-duplication of class names is a really nice feature, but it potentially makes this function somewhat expensive. I suspect there's room for further optimization here.

Two ideas:
1.   Use a stateful regex iteration within a while loop instead of using `String.prototype.split` which involves an array allocation. I suspect building the set of unique classes should be faster this way.

2. Use of a modified version of `injectStyle` from `styletron-utils` that returns an object with atomic class keys. Then merging can simply be repeated `Object.assign` and then an iteration of the final object keys to build the final class name string. I hadn't thought of this before, but this would definitely have an impact on https://github.com/rtsao/styletron/pull/91

wagerfield: I confess....the only reason I added this function was for test purposes 😁 . When composing styled components (`styled(AlreadyStyledComponent)`) I was seeing `className="a a b c"` which was inconsistent with React (I'm guessing React does class de-duplication?).

So I guess the question then becomes—"does it matter?". 

Do you know if there is an overhead in the browser when parsing an element's class list? Does having the same class repeated mean that the browser will parse and apply that style twice?

If it doesn't, then we can probably just drop it—I agree that it has a somewhat unnecessary cost to it. However, if it does have a browser parsing cost, then let's optimise it as you suggested above.

wagerfield:
Re. changing the class declaration name from `Provider` to `StyletronProvider`, I will do that now.

Inferno `1.3.0` stable version is being published today, so I will update the `package.json` deps to this as soon as it's released.

Re. the React updates—happy to take a look at porting the changes across so that the 2 packages are in close alignment.

Re. the generic test utils (`assign`, `isString`, `isNil` etc.) let me know what you want to do there and I/we can refactor them into the `styletron-core` or `styletron-utils` package. 

Super keen to get this PR merged ASAP because I really wanna be using Styletron in my Inferno projects 😄

packages/styletron-inferno/src/styled.js
wagerfield: Just to clarify, when I said "test purposes" I meant actually writing the unit tests.

faddee: @rtsao how do you get duplicated class names? The only way I know of is:

```jsx
const Root = styled('div', {color: 'red'});
const MyComponent = ({className}) => <Root className={className} />
module export default styled(MyComponent, {color: 'red'});
```

But that's not the case in the tests of what I can see.

rtsao: @faddee Yeah, you're right! The style objects get merged before being passed into injectStyle. Not sure what I was thinking before 😅. @wagerfield I think this function is basically entirely redundant.

faddee: What a relief! 😌 You almost scared me there.

.gitignore
```diff
@@ -4,4 +4,3 @@ npm-debug.log.*
 htmldocs/
 lerna-debug.log
 lib/
-.DS_Store
```
wagerfield: Why did you remove this? I see `.DS_Store` files in my untracked files all the time. I think this should be there prevent Mac users from inadvertently committing them—or save them from having to manually select the files they want to commit to avoid committing `.DS_Store` files.

rtsao: This shouldn't this be in the global `.gitignore`?

wagerfield: One way of tackling it, but personally I don't like using global git settings for anything par user name/email—I believe it's much safer to have this kind of stuff configured on a project by project basis. That way if someone works on the project and don't have global settings that ignore these system files (like me) they don't end up accidentally committing these files. Your repo, your call—just my 2 cents :)

wagerfield:
Ahead of publishing `styletron-inferno` to npm, I'd really like to update the `inferno` deps in `package.json` to the latest stable version—`1.3.0` (dropping the current release candidate status). This should be published in the coming days.

.gitignore
wagerfield: Put it another way—you shouldn't ever really commit `*.logs` or the `node_modules` folder, yet these files/folders are _always_ added to a project's `.gitignore` as opposed to assuming that they're in a user's global `.gitignore`.

Why should `node_modules` and `.log` files be treated any differently to system files like `.DS_Store`? 

Having some files/folders ignored globally assumes that everyone else that may work on the repo has the same global settings as you.

PR #91, opened 2017-02-21
Create `injectStyle` function with plugins and middlewares

### LINKED ISSUE #39: [styletron-utils] Support transformation middlewares
As an example - [jss](https://github.com/cssinjs/jss) has [plugins](https://github.com/cssinjs/jss/blob/master/docs/plugins.md) to modify styles before injecting them.

### My Proposal

Add `injectStyle` creator - that will accept `transformer` function as it's first argument

```javascript
export function createInjectStyle(transform) {
  return function injectStyle(styletron, styles, media, pseudo) {
     // transformations
  }
}
```

And first `injectStyle` with transfromer would be `injectStylePrefixed`:

```javascript
export const injectStylePrefixed = createInjectStyle(() => /* add prefixes */) 
``` 

Also every lib that use `styletron-utils` should have an option to pass custom `injectStyle` functions.

e.g. `styletron-react` :

```javascript
const injectStyles = createInjectStyle(compose(...myTransformers))

ReactDOM.render(
  <StyletronProvider
    injectStyles={injectStyles} 
    styletron={new Styletron(styleElements)}>
    <App/>
  </StyletronProvider>
);
```

faddee:
Basically replaces the whole `styletron-utils` package with the ability to provide plugins to manipulate styles before the declarations injects through middlewares.

~~Note that I took a little bit of a chance and did a change that might cause a major release. The major change is that you pass the `styletron` instance as only argument to `injectStyle` which returns a new function that accepts the other three (`style`, `media` and `pseudo`). The reason I made that choice:~~
- ~~`injectStyle` is now a required prop for `<StyletronProvider />` otherwise we force it to bundle even when it's not used~~
- ~~It felt unnecessary to pass both `styletron` and `injectStyle` as prop to `<StyletronProvider />` when the instance provides `injectStyle` with a single method and nothing else.~~ 
- ~~I believe it also feels more natural to reuse the function for other use cases: `module export injectStyle(styletron)`~~

~~I understand that there may be other than React users to take in consideration but if this may cause any trouble or you don't agree, just let me know.~~

Here's an example of a basic implementation:

```jsx
import StyletronClient from 'styletron-client';
import {StyletronProvider} from 'styletron-react';
import {injectStyle} from 'styletron-utils';

const styletron = new StyletronClient();
<StyletronProvider styletron={styletron} injectStyle={injectStyle}>{...}</StyletronProvider>
```

To make a custom `injectStyle` you simply pass a plugin or multiple composed plugins as first argument to `createInjectStyle`:

```jsx
import StyletronClient from 'styletron-client';
import {StyletronProvider} from 'styletron-react';
import {createInjectStyle} from 'styletron-utils';
import prefixAll from 'inline-style-prefixer';
import otherStylePlugin from 'third-party-style-plugin';

const compose = (...fns) => initial => fns.reduceRight((result, fn) => fn(result), initial);

const styletron = new StyletronClient();
const injectStyle = createInjectStyle(compose(prefixAll, otherStylePlugin));
<StyletronProvider styletron={styletron} injectStyle={injectStyle}>{...}</StyletronProvider>
```

This term middleware in this case are functions that handles array values and media- and psuedo properties. These are also overridable for customization. But if you define a middleware, the default ones won't be included and you will have to add those you want manually:

```js
const injectStyle = createInjectStyle(
  compose(prefixAll, otherStylePlugin),
  customMiddleware,
  fallbackValueMiddleware,
  mediaMiddleware,
  pseudoMiddleware
);
```

Fixes #39

~~EDIT~~

~~A featured "old" `<StyletronProvider />` with `injectStylePrefixed` as default is now accessible by importing `styletron-react/featured`. So that will be the tiniest effort `styletron-react` users will have to change after an upgrade. If someone doesn't change the import path or provide an `injetStyle` prop, it will throw.~~

~~The reason I choose this solution and not the contrary (a `injectStyle` provider from `styletron-react/custom`) is because you would in that case have to import `styletron-react/custom` everywhere if your bundler doesn't provide tree-shaking.~~

### DIFF
--- a/packages/styletron-react/src/provider.js
+++ b/packages/styletron-react/src/provider.js
@@ -1,5 +1,6 @@
 const React = require('react');
 const PropTypes = require('prop-types');
+const {injectStylePrefixed} = require('styletron-utils');
 
 /**
  * @class StyletronProvider
@@ -15,18 +16,21 @@ const PropTypes = require('prop-types');
  *     </StyletronProvider>
  *   );
  * }
- * 
+ *
  * @property {object} styletron - Styletron instance
  * @property {ReactElement} children - children
  * @extends ReactClass
  */
 class StyletronProvider extends React.Component {
   getChildContext() {
-    return {styletron: this.styletron};
+    return this.childContext;
   }
   constructor(props, context) {
     super(props, context);
-    this.styletron = props.styletron;
+    this.childContext = {
+      styletron: props.styletron || context.styletron,
+      injectStyle: props.injectStyle || context.injectStyle || injectStylePrefixed
+    };
   }
   render() {
     return React.Children.only(this.props.children);
@@ -34,12 +38,13 @@ class StyletronProvider extends React.Component {
 }
 
 StyletronProvider.PropTypes = {
-  styletron: PropTypes.object.isRequired,
-  children: PropTypes.element.isRequired
+  styletron: PropTypes.object,
+  children: PropTypes.element
 };
 
-StyletronProvider.childContextTypes = {
-  styletron: PropTypes.object.isRequired
+StyletronProvider.contextTypes = StyletronProvider.childContextTypes = {
+  styletron: PropTypes.object,
+  injectStyle: PropTypes.func
 };
 
 module.exports = StyletronProvider;
--- a/packages/styletron-react/index.d.ts
+++ b/packages/styletron-react/index.d.ts
@@ -1,9 +1,14 @@
 import * as React from "react";
 import StyletronServer from "styletron-server";
 import StyletronClient from "styletron-client";
+import {InjectStyle} from 'styletron-utils';
 
 declare namespace StyletronReact {
-  export class StyletronProvider extends React.Component<{styletron: StyletronServer | StyletronClient}, void> {}
+  type ProviderPropType = {
+    styletron?: StyletronServer | StyletronClient;
+    injectStyle?: InjectStyle;
+  };
+  export class StyletronProvider extends React.Component<ProviderPropType, void> {}
 
   type InnerRef<TInstance> = {
     innerRef?: (instance: TInstance) => any;
--- a/packages/styletron-react/src/styled.js
+++ b/packages/styletron-react/src/styled.js
@@ -1,6 +1,5 @@
 const React = require('react');
 const PropTypes = require('prop-types');
-const utils = require('styletron-utils');
 
 const isValidAttr = require('./is-valid-attr');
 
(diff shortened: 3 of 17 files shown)


faddee:
One idea to prevent a major change could be to move this `<StyletronProvider />` to `<CustomStyletronProvider />` and revert `<StyletronProvider />` to accept the `styletron` prop which will use the new `injectStylePrefixed` as default. In that way bundlers will at least have a chance to tree-shake `injectStylePrefixed` if it's not used.

rtsao:
Nice! Middleware is definitely a nice feature.

Is there a use case for having different middlewares in different parts of your app?

If the instance and `injectStyle` remained decoupled, you could do do something like:

```jsx
<StyletronProvider instance={styletron} injectStyle={baseInject}>
  <Stuff/>
  <OtherStuff>
    <StyletronProvider injectStyle={specialInject}>
      <MoreStuff/>
    </StyletronProvider>
  </OtherStuff>
</StyletronProvider>
```

I also feel like if the core instance is provided, this leaves more room for flexibility as more core API features get added. Indeed, as it stands everything goes through `injectStyle`. However, I can imagine additional methods being added for example:

- `prefetchFontFace`
  - Don't inject the font face into the critical CSS, but mark the font to be asynchronously loaded (unless already injected)

There's a lot I want to do with font loading and declarative performance optimization. I'm not sure if that'll end up in Styletron or something else that uses/composes Styletron, but there's definitely the possibility of more methods beyond just `injectStyle`.

faddee:
Decoupled the instance and `injectStyle`. See edit in description.

packages/styletron-react/src/provider.js
```diff
@@ -1,44 +1,18 @@
...
-  getChildContext() {
-    return {styletron: this.styletron};
+module.exports = createProvider(
+  ({styletron, injectStyle}) => {
+    if (!injectStyle && styletron) {
+      throw new Error('You need to switch to the exported `StyletronProvider` from `styletron-react/featured` in this upgraded version. For more info please visit: https://github.com/rtsao/styletron/');
+    }
+    return {
+      styletron,
+      injectStyle
```
faddee: Realized that other libs expects the instance to be on `context.styletron`. Need to find a better way to pass `injectStyle` without adding a new property to `context`.

faddee: Can find a way of doing this without cloning the styletron instance which I don't think would be preferred. Created two properties on context anyway.

faddee:
The `injectStyle` function is now overridable by the `<StyletronProvider />`.

collin:
@faddee do you understand #97 enough to comment on whether this PR would cover our use?

faddee:
It's been two months. What's your thoughts? Is there any intention of merging this?

faddee:
Merge conflicts solved. Also reverted the part where there was two entry points for importing `styletron-react` since my gut feeling tells it's where the shoe pinches.

faddee:
Thanks for your feedback. Your alternative approach is better. Letting the user create it's own `styled`-functions. Closing this PR in favor for a new one.

PR #94, opened 2017-02-28
Remove redundant class dedupe function from styletron-inferno

rtsao:
Per the discussion in https://github.com/rtsao/styletron/pull/90, the added class dedupe function seems to be redundant.

cc @faddee @wagerfield

### DIFF
--- a/packages/styletron-inferno/src/styled.js
+++ b/packages/styletron-inferno/src/styled.js
@@ -72,7 +72,11 @@ function createStyledComponent(name, stylesArray) {
     const newProps = assign({}, props);
     const styles = resolveStyles(stylesArray, props, context);
     const className = injectStylePrefixed(context.styletron, styles);
-    newProps.className = mergeClassNames(props.className, className);
+
+    newProps.className = props.className
+      ? `${props.className} ${className}`
+      : className;
+
     if (isFunction(props.innerRef)) {
       newProps.ref = props.innerRef;
       delete newProps.innerRef;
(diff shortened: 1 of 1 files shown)


wagerfield:
Could very well be, though I'm 99% sure I was seeing an element with a className of `a a b c` in one of my early tests, so let me take a look and see if I can repeat this. Either way since each class just maps to a single CSS property/value declaration, it doesn't actually matter from a styled result perspective—the reason for doing it was more to save the browser from potentially parsing and applying the class twice. Given your knowledge on how browsers process classes, and CSS—do you see this being an issue?

rtsao:
Yeah, I'd love to add a test case for duplicated class names.

faddee:
It's not a duplicated `{color: 'red'}` that would worry me. It's if the composition of `{color: 'red'}` and `{color: 'green'}` doesn't override the other that would worry me. But the tests are running just fine now and having `styletron-react` and `styletron-inferno` as identical as possible will ease all kind of maintenance. So I encourage to merge this.

wagerfield:
No idea what I did before to get duplicate class names—I expect it was an error on my part in an earlier implementation. Either way, I am unable to reproduce this test case, so happy for you to merge this 👍 

@faddee I agree—the closer these implementations are, the better.

Could either of you please explain _why_ you need to check for invalid attributes in the React `styled` implementation (`omitInvalidProps > isValidAttr`)?

faddee:
@wagerfield Because React warns in development mode when you pass [unknown props](https://facebook.github.io/react/warnings/unknown-prop.html) to an element because they consider it to be an anti-pattern. But React still omits unknown props so the checks could theoretically be disabled in production.

wagerfield:
But is that not then the responsibility of the user using React as opposed to the responsibility of Styletron to 'cleanse' the user's errors? 

Like if I used React without Styletron and added some unknown props to an element, I would get those warnings right? Then I would go "oh, my bad" and then fix them. 

I don't think this should be the responsibility of Styletron since it's of no concern to applying styles _and_ it adds unnecessary bloat and overhead to the lib. I say drop it 😄

vikfroberg:
@wagerfield you can pass a function to `styled` which makes it possible to produce different styles depending on which props you send to the component. There's no way for the user to "clean" those props before they're sent down to the React element, and so that's the reason Styletron does that for you :)

```jsx
const Text = styled('span', props => ({
  fontSize: props.small ? '12px' : '16px',
}))

<Text small>Hello world!</Text>
```

wagerfield:
Of course—hadn't thought of that, thanks @vikfroberg 👍

PR #97, opened 2017-03-02
proof of concept for "readable" class names

moimikey:
hola,

styletron is wonderful and at my company, we've been doing some internal demos as a potential option for introducing it into our codebase. some engineers though are concerned about the way the class names are named (they're simply hashes) so the readability goes out the door. this PoC introduces a new constructor option called `readableClassNames` (obviously this can be subject to rename), which will instead use a combination of `prefix + prop + val`, joined by hyphens, or `prop + val`, instead of letter hashes.

the naming concat would need to be a little more elegant, as you'll see in the screenshot where I'm rendering it in `react-storybook`

> without prefix

screen shot 2017-03-02 at 2 06 33 pm

> with prefix

screen shot 2017-03-02 at 2 12 43 pm

let me know your thoughts...

### DIFF
--- a/packages/styletron-core/src/index.js
+++ b/packages/styletron-core/src/index.js
@@ -6,15 +6,17 @@ class StyletronCore {
 
   /**
    * Create a new StyletronCore instance
-   * @param {object} [opts]           An object containing options
-   * @param {string} [opts.prefix=''] A prefix for generated CSS class names
+   * @param {object} [opts]                           An object containing options
+   * @param {string} [opts.prefix='']                 A prefix for generated CSS class names
+   * @param {boolean} [opts.readableClassNames=false] Use readable class names
    */
-  constructor({prefix = ''} = {}) {
+  constructor({prefix = '', readableClassNames = false} = {}) {
     this.cache = {
       media: {},
       pseudo: {}
     };
     this.prefix = prefix === '' ? false : prefix;
+    this.readableClassNames = readableClassNames;
     this.uniqueCount = 0;
     this.offset = 10; // skip 0-9
     this.msb = 35;
@@ -59,7 +61,8 @@ class StyletronCore {
       return cached;
     }
     const virtualCount = this.incrementVirtualCount();
-    const hash = virtualCount.toString(36);
+    const readableClassName = decl.prop + '-' + decl.val;
+    const hash = this.readableClassNames ? readableClassName : virtualCount.toString(36);
     const className = this.prefix ? this.prefix + hash : hash;
     StyletronCore.assignDecl(this.cache, decl, className);
     return className;
--- a/packages/styletron-core/src/test/index.js
+++ b/packages/styletron-core/src/test/index.js
@@ -56,3 +56,20 @@ test('test injection with prefix', t => {
   t.equal(instance.getCount(), 1, 'unique count incremented');
   t.end();
 });
+
+test('test injection with readable class names', t => {
+  const instance = new StyletronTest({readableClassNames: true});
+  t.equal(instance.readableClassNames, true, 'readableClassNames is set on instance');
+  t.equal(instance.getCount(), 0, 'starts with 0 declarations');
+  const decl1 = {prop: 'color', val: 'red'};
+  instance.injectDeclaration(decl1);
+  t.equal(instance.getCache().color.red, 'color-red');
+  t.equal(instance.getCachedDeclaration(decl1), 'color-red');
+  t.equal(instance.getCount(), 1, 'unique count incremented');
+  const decl2 = {prop: 'backgroundColor', val: 'orange'};
+  instance.injectDeclaration(decl2);
+  t.equal(instance.getCache().backgroundColor.orange, 'backgroundColor-orange');
+  t.equal(instance.getCachedDeclaration(decl2), 'backgroundColor-orange');
+  t.equal(instance.getCount(), 2, 'ends with 2 unique declarations');
+  t.end();
+});
--- a/packages/styletron-react/src/test/browser.js
+++ b/packages/styletron-react/src/test/browser.js
@@ -105,6 +105,20 @@ test('styled component', t => {
   t.end();
 });
 
+test('styled component with readable class names', t => {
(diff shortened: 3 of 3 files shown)


collin:
Hi, I can provide some extra background here. We're evaluating styletron and possible building some extra tooling on top of it.

Rather than pass a plain-old object like this:

```js
styled('div', {
  background: 'blue',
  width: '20px',
})
```

We're passing in instances of classes in as the values:
```js
const PrimaryBrandColor = new ColorVariable('blue');
const MediumSize = new SizeVariable('20px'); 

styled('div', {
  background: PrimaryBrandColor,
  width: MediumSize
})
```

To get this working we actually have defined our own version of the styled function called `css` that  analyzes/stash the variables and creates view into our styles structure before passing the string version through to `styled`.

What would be very helpful for our efforts would be hook points in to do two things:

1. Specify our own custom class names to assuage developers during development.
2. Convert our variable classes into strings, or expand them into a group of properties.

re: groups of properties, we are considering using virtual properties such as `layout` or `font` to map over multiple css props.

```js
css('div', {
  layout: HorizontalBaselineFlexboxLayout,
  font: ExtraLargeSansSerif,
})
```

Which is equivalent to:

```js
styled('div', {
  display: 'flex',
  flexDirection: 'row',
  alignContent: 'baseline',
  fontFamily: 'sans-serif',
  fontSize: '8rem',
})
```

collin:
This is a pretty raw POC: but for reference the variable classes we're using right now look like this:

https://gist.github.com/collin/5aa586cafc9e06c0bbf82a47cbcfeb49

collin:
Hm, we might be covered #91

faddee:
No, you can't manipulate Styletron class names with #91 if that's what you are asking?

collin:
Yes, that answers the question, thank you.

rtsao:
This is a cool idea! Agreed this can could make DX better (although I suppose the class names can get rather long in some cases). I assume you would only have this mode in development?

I eventually want to have a full-fledged Chrome DevTools integration with a styles inspector that alleviates the need for this (and hopefully provide a workaround for https://bugs.chromium.org/p/chromium/issues/detail?id=387952 as well). However, in the meantime this is pretty good.

I'm also very concerned about bundle size, but this change only adds a few bytes so I'm more comfortable rolling this into the core. But ideally I think the `readableClassNames` code is completely eliminated in a production build which would also eliminate the runtime cost of the conditional in `injectDeclaration`. But need to do some refactoring of how packages are published (I need to create rollup builds) before we get to that, so for now I think this is fine.

packages/styletron-core/src/index.js
```diff
@@ -59,7 +61,8 @@ class StyletronCore {
       return cached;
     }
     const virtualCount = this.incrementVirtualCount();
-    const hash = virtualCount.toString(36);
+    const readableClassName = decl.prop + '-' + decl.val;
```
rtsao: I think placing this in a nested ternary would be preferable here so we avoid performing unnecessary string concat

packages/styletron-core/src/index.js
```diff
@@ -6,15 +6,17 @@ class StyletronCore {
...
   /**
    * Create a new StyletronCore instance
-   * @param {object} [opts]           An object containing options
-   * @param {string} [opts.prefix=''] A prefix for generated CSS class names
+   * @param {object} [opts]                           An object containing options
+   * @param {string} [opts.prefix='']                 A prefix for generated CSS class names
+   * @param {boolean} [opts.readableClassNames=false] Use readable class names
    */
-  constructor({prefix = ''} = {}) {
+  constructor({prefix = '', readableClassNames = false} = {}) {
```
rtsao: What about something like `compressedClassNames = true`. Might be more clear than trying to describe the "readable" class name form.

rtsao: I'm open to other thoughts/suggestions as well!

rtsao: Or maybe `sequentialClassNames`

moimikey:
i'll update this PR tomorrow AM :) thanks for taking the time to look it over; also, totally agree that devtools integration would be useful, and yes, it's primarily/solely for development purposes.

packages/styletron-core/src/index.js
jbellsey: A late preference is expressed: `expandClassNames = false` by default, or `expandClassNames = true` as a manual override during development.

I prefer the command form ("expand" as opposed to "expanded"). And I'm not fond of "readable", since they're only marginally readable :) 

Also, the semantics of compression-by-default is important. I'd prefer to steer clear of anything that names the behavior in any way other than the default (e.g., "sequential" or "compressed"). 

Carry on, and I love the idea! Will use during development for sure.

packages/styletron-core/src/index.js
rtsao: I've just realized this, but this mode will produce collisions that won't occur with the standard, sequential class names.

This will happen with media queries and pseudo classes, for example:

```js
styletron.injectDeclaration({color: 'red'});
// → 'a' (standard mode)
// → 'color-red' 
styletron.injectDeclaration({color: 'red', media: 'screen'});
// → 'b' (standard mode)
// → 'color-red'
styletron.injectDeclaration({color: 'red', pseudo: ':hover'});
// → 'c' (standard mode)
// → 'color-red'
```

A couple of possibilities:
1) *Always* use standard, sequential class names in cases of declarations with `psuedo` and/or `media` attributes.
2) Have additional prefixes based on `psuedo` or `media` properties. The trick here is we need a bijection from these values to valid class name identifiers.

ehellman:
Would this be an option or overall change? Because I think that one of the great features of Styletron is the small size of the HTML that you get from using it, helps us a ton with performance on SSR. I could understand the need for readable classNames in dev maybe, if you dont use React DevTools (then you can use `displayName`).

rtsao:
The codebase has changed quite a bit since this PR was opened. I think adding this functionality into `styletron-engine-atomic` is an interesting idea (in development?)

For readability I recommend using the new debug mode: https://github.com/styletron/styletron/tree/master/packages/styletron-react#browser-debug-mode

Issue #92, opened 2017-02-23
[RFC] @keyframes and @font-face proposal


rtsao:
I'm still not sure about the API yet, but I was thinking it might look roughly something like the following:

### Additional instance methods on the core `Styletron` class
- `constructor()`
- `injectDeclaration(decl : object) -> string`
- `injectFontFace(font : object) -> string` (new)
  - returns name of injected font
- `injectKeyframes(keyframes : object) -> string` (new)
  - returns name of injected keyframe animation

**Core vs mixins**
Maybe these are in the core or maybe they are mixins. Probably some minor internal refactoring needed either way. Need to think about additional features that might be desired to see if a mixin interface is warranted.

**With media queries?**
There needs to be more investigation into nesting`@font-face` and `@keyframes` inside of media queries as this will affect the interface. The spec seems to disallow this, but apparently some browser support it?

**Naming/scoping?**
Should there be a way to manually set animation/keyframe names? I think they should just be sequential (`a`, `b`, `c`, etc) with an optional namespace (just like atomic declarations).

**Caching**
For declarations, there's only 4 properties with string values required to check equality. @font-face rules and @keyframes rules are much trickier. I think simple reference equality might be sufficient here, as typically these are defined separately. See below for why reference equality is probably OK in my opinion, as it's reasonable to expect these are hoisted outside of render.

### New primitives for `styletron-react` and friends

```js
// Note: just an idea...
const {styled, fontFace, keyframes} from 'styletron-react';

const myFont = fontFace({...fontProperties}); // no side effect
const myAnimation = keyframes({...keyframeProperties}); // no side effect

const Box = styled('div', {
  fontFamily: myFont, // injects once on first render 
  animationName: myAnimation // injects once on first render
});
```

jbellsey:
Great to see movement on this!

I don’t love the idea of adding new magic to `styletron-react`, since we’ve written our own HoC to replace `styled`. I’d prefer having these features in one of the core libs.

I also wonder if tracking the new classes (e.g., `myFont` or `myAnimation`) is a feature that ought to be provided by styletron, rather than the app. These features are global by design. So one way or another, the application needs to give components a global cache, either on context, through a global store, or (see below) directly managed by styletron. These classes have zero likelihood of being manipulated in code -- they're set-and-forget -- which is why they might make sense to track in styletron:

```js
import {addFontFace, retrieveFontFace} from 'styletron-utils';

// at startup:
addFontFace(‘icons’, {...fontProperties});

// then:
[...]

PR #105, opened 2017-03-22
Added basic support for keyframe animations

andreiduca:
Ref. #92

**NOTE:** This is still in development. While it does work, it has some drawbacks. Please read below. Thanks!

### Changes

- Added a new key to the cache object called `keyframes`. This object will hold our animations.
- Since keyframe animations cannot be combined as normal classes are, the keys of this new object are the json stringified versions of the animation objects, while the values represent the generated animation names. This way, identical animation objects will have the same generated name.
- A class name is also added with the new animation-name. To keep things simple, the animation name and the class name are the same.

### Usage:

```js
const Box = styled('div', {
    animationName: {
        '0%': {
            'margin-top': 0
        },
        '100%': {
            'margin-top': '100px'
        }
    }
});
```

### Known issues

- CamelCase support is not fully implemented; use `'margin-top'` rules instead of `maginTop`
- Shorthand `animation` is not implemented. This will require a helper function that would retrieve an animation name based on a given rule set object.
- String value for animationName is not supported.
- Vendor prefixes for `@keyframes {...}` are not implemented.
- Theoretically, this should also work with media queries. But I haven't tested it yet.

### DIFF
--- a/packages/styletron-client-ie9/src/__tests__/browser.js
+++ b/packages/styletron-client-ie9/src/__tests__/browser.js
@@ -24,7 +24,7 @@ class StyletronTest extends Styletron {
 test('hydration basic', t => {
   const elements = createFixtures([
     {
-      css: '.e:hover{display:none}.a{color:red}.b{color:green}',
+      css: '@keyframes f{0%{border-bottom:1px;margin-left:0}100%{border-bottom:10px;margin-left:10px}}.e:hover{display:none}.a{color:red}.b{color:green}.f{animation-name:f}',
     },
     {
       media: '(max-width: 800px)',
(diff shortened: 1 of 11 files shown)


packages/styletron-client/src/styletron-client.js
```diff
@@ -1,4 +1,5 @@
-const DECL_REGEX = /.([^:{]+)(:[^{]+)?{([^:]+):([^}]+)}/g;
+const DECL_REGEX = /\.([^:{]+)(:[^{]+)?{([^:]+):([^}]+)}/g;
```
andreiduca: Here i just escaped the leading **.** (dot)

andreiduca:
I believe this is closed in #217 .

Issue #98, opened 2017-03-02
Publish rollup builds


rtsao:
- [ ] Convert from CommonJS to ES Modules everywhere
- [ ] Generate CommonJS and ES Module bundles with Rollup (also UMD?)

This will make tree shaking possible and solve a lot of issues with bloat and styletron-utils

ludovicofischer:
I could give this a shot. What’s the main reason you would like ES6 modules? Step 1 could be partly achieved with https://github.com/nolanlawson/cjs-to-es6

rtsao:
Two reasons:

1. We can publish "prebundled" modules with Rollup, which reduces [module overhead](https://nolanlawson.com/2016/08/15/the-cost-of-small-modules/) for consumers. So each package will just be a single file/module, rather than possibly multiple files/modules.
2. styletron-utils has lots of helper functions and whatnot, but if it is a CommonJS package, most bundles cannot tree shake unused portions. If a ES Module is also provided, webpack will be able to tree shake.

ludovicofischer:
Hi, I’ve experimented with rollup builds (https://github.com/rtsao/styletron/compare/master...ludovicofischer:master).

Would you run the tests against the bundles or leave them as now?

Since we would need to use [rollup-plugin-commonjs](https://github.com/rollup/rollup-plugin-commonjs) for external dependencies anyway, it might be best to test the rollup build does not somehow break the published artifacts before converting the styletron code to ES6 modules.

ludovicofischer:
I’d definitely say move to rollup and only _after_ convert to ES6 modules.

Since [bublé can’t transpile modules](https://buble.surge.sh/guide/#using-es-modules), if we concert the source to ES6 modules, then either the tests must use the bundled library or we must switch to Babel. Using the bundled library might not be ideal if you want to test functions that are not exported by the final bundle, but switching to Babel should be handled with some care.

On the other hand, just publishing a bundle should reduce module overhead. And solve https://github.com/rtsao/styletron/issues/77 since adding UMD should just be an additional line in the rollup config.

PR #106, opened 2017-03-22
Build modules with rollup

ludovicofischer:
Part of #98 . I’ve left styletron and styletron singleton alone as I was not sure it was worthwhile to convert them. I also added a UMD bundle for styletron client and styletron-client-ie9.

I still need to test that importing the modules in a project works. Maybe in the end it would make sense to test the bundles, as I did not find tests that exercise private modules.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,6 @@ npm-debug.log
 npm-debug.log.*
 htmldocs/
 lerna-debug.log
+yarn-error.log
 lib/
+dist/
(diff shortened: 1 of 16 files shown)


rtsao:
So I ended up doing a comprehensive, total overhaul and refactor of the codebase in https://github.com/rtsao/styletron/pull/125 including migrating to ES2015 modules, which makes working with Rollup a bit easier and more robust. As part of this, I also factored out what I consider to be best practices into https://github.com/rtsao/create-universal-package which this repo will use going forward (also helping to avoid duplication of config everywhere). I do really appreciate your effort, but https://github.com/rtsao/styletron/issues/98 ended up being an exceeding large undertaking that ended up being much larger in scope than I had originally envisioned.

PR #111, opened 2017-04-05
[WIP] Styletron Chrome DevTools extension

rtsao:
This PR contains a WIP proof of concept for a Styletron Chrome DevTools extension.

It is currently a very basic sidebar pane that is added to the Elements panel. When you select an element that is a `styletron-react` styled element, the resolved Styletron style object will appear.

screen shot 2017-04-04 at 6 05 00 pm

I'm open to suggestions on what this should look like and which features there should be. Please comment with your ideas! The sidebar pane can contain arbitrary HTML/CSS/JS so it can be anything. Or it could even be a full panel (like the React DevTools extension).

A few initial ideas:
- When a styled element extends another styled element, show both the base style object and the overriding style object.
- Editing of style objects. This will probably be somewhat challenging.
- Also display a mapping of declarations to atomic class names

### DIFF
--- a/packages/styletron-react/src/styled.js
+++ b/packages/styletron-react/src/styled.js
@@ -6,6 +6,18 @@ const isValidAttr = require('./is-valid-attr');
 const STYLES_KEY = '__STYLETRON_STYLES';
 const TAG_KEY = '__STYLETRON_TAG';
 
+if (typeof window !== 'undefined' && !window.__STYLETRON_DEVTOOLS__) {
+  const stylesMap = new WeakMap();
+  window.__STYLETRON_DEVTOOLS__ = {
+    stylesMap,
+    getStyles: (element) => {
+      if (stylesMap.has(element)) {
+        return stylesMap.get(element);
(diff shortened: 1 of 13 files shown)


packages/styletron-react/src/styled.js
```diff
@@ -6,6 +6,18 @@ const isValidAttr = require('./is-valid-attr');
 const STYLES_KEY = '__STYLETRON_STYLES';
 const TAG_KEY = '__STYLETRON_TAG';
 
+if (typeof window !== 'undefined' && !window.__STYLETRON_DEVTOOLS__) {
```
faddee: Maybe these new blocks should be wrapped with `if (PROCESS.ENV.NODE_ENV !== 'production') { ... }`?

Primajin:
Cool! Could you provide details on how to implement this "beta version" into chrome until it is merged and released?

tajo:
closed because of https://github.com/styletron/styletron/pull/305

PR #118, opened 2017-04-18
Move `prop-types` to `dependencies`

faddee:
Per [recommendation by `prop-types`](https://github.com/reactjs/prop-types/blob/master/README.md#how-to-depend-on-this-package) it should be a dependency and not a peer dependency.


rtsao:
Nice catch, thanks!

PR #120, opened 2017-04-19
Intersect `CSSProperties` instead of extending

### LINKED ISSUE #99: Typescript definition styletron-react performance
After using this package for around a week my Visual Studio Code hints and error checking has slowed considerably (sometimes up to 10 second delay on error or type information). I believe I've tracked it down to my use of styletron-react.

I think it's something to do with the way generics are used in the type definition, since when I run `tsc` with the `--diagnostics` flag, I have a huge number of types.

- ~400 types: `const Component2 = styled(Component1, {});`
- ~50 types: `const Component2 = styled('div', {});`
- ~1200 types `const Component2 = styled(Component1, ({ }) => ({}));`

Not sure if typescript version matters, but I've had the issue on 2.2 and 2.1. Also don't think .tsconfig matters since I got these numbers from a single file.

@faddee I see you did a lot of work on the typings, have you run into any similar issues?

In the meantime I've added a function in front of the `styled` function to strip any typing and my project is now fast again.

```typescript
(base: any, style: any) => styled(base, style) as any;
```

I'll keep investigating and see if I can get those numbers down.

faddee:
It appears that this provides a significant performance improvement when type checking for some reason and will hopefully fix #99

### DIFF
--- a/packages/styletron-react/index.d.ts
+++ b/packages/styletron-react/index.d.ts
@@ -9,12 +9,9 @@ declare namespace StyletronReact {
     innerRef?: (instance: TInstance) => any;
   };
 
-  // This is not entierly true, but extending `CSSProperties` enables
-  // autocompletion for CSS properties as plain objects and not only
-  // for function return objects
-  interface StyleFunction<TProps> extends React.CSSProperties {
-    (props: TProps): React.CSSProperties;
-  }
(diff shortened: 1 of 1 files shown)


rtsao:
Sounds good! 👍

PR #124, opened 2017-05-02
TypeScript typing tests and reflect the return type of `styled` as stateless component

faddee:
Thanks to https://github.com/rtsao/styletron/commit/0ec7c18829053b59db387c6d0674f8407a736250 we can remove the hack for some reason and extend the official `React.StatelessComponent` which gives the users a truthful return type of `styled`.

I also added a simple typings test.


rtsao:
Nice 😄!

PR #127, opened 2017-05-09
Add `core` function

faddee:
Moved basically the whole `styled` to `core` instead and added two new arguments:
 - `mapStyleToProps`: Function that consumes the style object result and returns an object with class names
    
    ```ts
    function mapStyleToProps(styletron: Styletron, style: object, props?: object): object
    ```
 - `mergeProps`: Optional, a function where you can override the merging of props and the returned object from `mapStyleToProps`
    
    ```ts
    function mergeProps(styleProps: object, ownProps: object): object
    ```

The reason for this to provide the users the ability to create their custom `styled` function where they can manipulate style and props during render.

See history behind this PR: https://github.com/rtsao/styletron/pull/91#pullrequestreview-34732010

TypeScript now requires 2.1 or greater in favor of the `keyof` operator and mapped types.


faddee:
cc @rtsao I believe I'm quite ready. Sorry for the WIP. But that's the only way for me as Windows user to run your tests. :-) Any thoughts on this take?

faddee:
I havn't tried but theoretically this should be all that's needed for a `withStyle` HOC:

```jsx
function withStyle(styles) {
  return component => {
    return core(
      component,
      styles,
      (styletron, styleResult, ownProps) => ({
        classes: Object.assign(...Object.keys(styleResult).map(name => ({
          [name]: injectStylePrefixed(styletron, styleResult[name]),
        }))),
      })
    );
  }
}

@withStyle({
  foo: {color: 'red'},
  bar: {color: 'blue'},
})
class MyComponent extends React.Component {
  render() {
    return (
      <div>
        <span className={props.classes.foo} />
        <span className={props.classes.bar} />
      </div>
    );
  }
}
```

rtsao:
Nice work, this is excellent! Sorry for the delayed response, I've been on vacation and without internet access 😄.

A few thoughts:

I think the `mapStyleToProps` function is generally going to be impure as it will most likely be calling imperative methods on the provided styletron instance. Additionally, since `ownProps` is provided as an argument to `mapStyleToProps`, there isn't really anything that stops someone from just modifying `ownProps` directly. With that in mind, would it make sense to combine the `mergeProps` and `mapStyleToProps` functions?

As a downside, there's some additional boilerplate (having to implement merging yourself), but I think for a low-level API this might offer the most flexibility. I can envision scenarios where the `mapStyleToProps` behavior and how prop merging behavior might need to be tightly coupled -- in which case this coupling would be more neatly packaged up as a single function. Also, I think that if the internal fast `assign` helper function were to be exported, it would make this boilerplate a bit more palatable.

As a plus, we'd be free to implement the same separated `mapStyleToProps` and `mergeProps` interface on top if we'd like.

faddee:
Nice with an internetless vacation. Sounds healthy. ☺️

Merging props in `mapStyleToProps` was my initial idea. But then I realized the extra boilerplate. So I looked over the api for the `connect` function for `react-redux`, which has a similar purpose. But this reminds me that their `mapStyleToProps` is generally an local function. In our case `mapStyleToProps` will likely be either an external function or coupled into an custom styling function. So I guess that makes the extra boilerplate ok in that point of view.

faddee:
I've smashed `mapStyleToProps` and `mergeProps` into one `assignProps` argument now. I don't export `assign` because it isn't faster than `Object.assign` and, in a React-world, I think we can assume that all users has their own way of assiging.

PR #129, opened 2017-05-16
[WIP] `withStyle` HOC

### LINKED ISSUE #33: [styletron-react] - add higher-order component to inject styles
Greetings! yesterday saw blog post about `styletron` and I'm superexcited  about it, wanted to test it in current project but found that it doesn't have hoc for React. 

That would be great to have HOC like this: 

```javascript
import {withStyles} from 'styletron-react';

const enhancer = withStyles((props) => ({
  foo: {
      color: props.alert ? 'orange' : 'lightblue',
  },
  bar: {
     backgroundColor: props.alert ? 'orange' : 'lightblue',
  },
}))

const Baz = (props) => {
  return (
    <div className={props.classes.foo}>
      <div className={props.classes.bar}>Baz</div>
    </div>
  )
}

export default enhancer(Baz)
```

frenic:
Fixes #33.

A `withStyle` HOC function is highly requested and will become more natural to use with React v16 whereof it will be possible to render an array of children.

Today it's only possible to pass `className` to one element using `styled`:

```jsx
function MyComponent(props) {
  return <div className={props.className} />
}

const Pretty = styled(MyComponent, {color: 'red'});
```

Doing this with React16 will become weird. So this PR provides a `withStyle` HOC function:

```jsx
function MyComponent(props) {
  return [
    <div className={props.classes.a} />,
    <div className={props.classes.b} />,
    <div className={props.classes.c} />
  ]
}

const Pretty = withStyle({
  a: {color: 'red'},
  b: {color: 'green'},
  c: {color: 'blue'},
})(MyComponent);
```

## Problem

With `styled`, it's possible to compose a styled component:

```jsx
function MyComponent(props) {
  return <div className={props.className} />
}

const Pretty = styled(MyComponent, {color: 'red'});

const Composed = styled(Pretty , {background: 'black'});

/* Result:
{
  color: 'red',
  background: 'black'
}
```

The question here is how composition should be handled for `withStyle`.

```jsx
function MyComponent(props) {
  return [
    <div className={props.classes.a} />,
    <div className={props.classes.b} />,
    <div className={props.classes.c} />
  ]
}

const Pretty = withStyle({
  a: {color: 'red'},
  b: {color: 'green'},
  c: {color: 'blue'},
})(MyComponent);

const Composed = withStyle({
  b: {background: 'black'},
})(Pretty);

/* Result 1:
{
  a: {color: 'red'},
  b: {
    color: 'green',
    background: 'black'
  },
  c: {color: 'blue'},
} */

/* Result 2 (current):
{
  a: {color: 'red'},
  b: {background: 'black'},
  c: {color: 'blue'},
} */
```

Since `core` doesn't deep merge styles, it will override `b` with the new style. **If "Result 1" is to be expected, which is the best way to do that?** A simple solution would be to "just" deep merge styles, which would also solve `@media` and `:hover` overrides. But I guess there's a good reason it's not implemented in the first place:

```jsx
const Pretty = styled('div', {
  '@media (min-width: 768px)': {color: 'red'}
});

const Composed = styled(Pretty , {
  '@media (min-width: 768px)': {background: 'black'} // Will override `{color: 'red'}`
});
```

### DIFF
--- a/packages/styletron-react/index.d.ts
+++ b/packages/styletron-react/index.d.ts
@@ -10,53 +10,66 @@ declare namespace StyletronReact {
 
   export function styled<TProps>(
     base: React.StatelessComponent<TProps & ClassNameProp>,
-    style: Style<TProps>): Component<TProps>;
+    style: Style<TProps, React.CSSProperties>): Component<TProps>;
 
   export function styled<TInnerProps, TOuterProps>(
     base: React.StatelessComponent<TInnerProps & ClassNameProp>,
-    style: Style<TOuterProps>): Component<TOuterProps & TInnerProps>;
+    style: Style<TOuterProps, React.CSSProperties>): Component<TOuterProps & TInnerProps>;
 
   export function styled<TProps, TComponent extends React.Component<TProps, React.ComponentState>>(
     base: React.ClassType<TProps, TComponent, React.ComponentClass<TProps & ClassNameProp>>,
-    style: Style<TProps>): Component<ClassProps<TProps, TComponent>>;
+    style: Style<TProps, React.CSSProperties>): Component<ClassProps<TProps, TComponent>>;
 
   export function styled<TProps, TBase extends keyof BasePropsMap<TProps>>(
     base: TBase,
-    style: Style<TProps>): Component<BasePropsMap<TProps>[TBase]>;
+    style: Style<TProps, React.CSSProperties>): Component<BasePropsMap<TProps>[TBase]>;
 
   export function styled<TElement extends HTMLElement, TProps>(
     base: string,
-    style: Style<TProps>): Component<HTMLProps<TProps, TElement>>;
+    style: Style<TProps, React.CSSProperties>): Component<HTMLProps<TProps, TElement>>;
 
   export function styled<TElement extends SVGElement, TProps>(
     base: string,
-    style: Style<TProps>): Component<SVGProps<TProps, TElement>>;
+    style: Style<TProps, React.CSSProperties>): Component<SVGProps<TProps, TElement>>;
 
   type ClassNameProp = {className?: string};
 
+  export function withStyle<TStyles extends MultiStyle>(styles: StyleFunction<{}, TStyles>): Decorator<TStyles>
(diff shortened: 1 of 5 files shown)


faddee:
cc @rtsao I need your input in this please. First of if this even is a good idea. Secondly how composition should be handled (see description). :-)

rtsao:
I decided against it initially for simplicity, but I had thought of building a multi-HOC solution with explicit merging. I think this is the most robust approach, since the user can explicitly opt-into whatever merge/transform behavior they wish.

I was thinking about something akin to this:

**`styled` (`tag` : string, `styles` : function | object) => StyledElementComponent**

- No more subclassing behavior by default, only creates new styled element components.

**`transformStyles` (`base` : StyledElementComponent, `transform` : function) => StyledElementComponent**

- The low-level composition primitive. Can be used for arbitrary transformation of styles.
- Perhaps this transform function will take the resolved styles and props as arguments?

Then on top of `transformStyles`, more user-friendly composition HOCs could be implemented:

**`withStyles` (`base` : StyledElementComponent, `styles` : function | object) => StyledElementComponent**

- Shallowly merges styles, but taking into account nested objects from media queries and pseudo selectors.

**`withStylesDeep` (`base` : StyledElementComponent, `styles` : function | object) => StyledElementComponent**

- Recursively deeply merges style objects

I think there needs to be some tweaking of the API, that's the general idea.

I like the idea of eagerly resolving/merging static styles, for example:

```js
const A = styled('div', {color: 'red'});
const B = withStyles(A, {color: 'blue'});
const C = withStyles(B, {color: 'green'});
```

Because the styles of A and B are static objects, for C, we could eagerly flatten the transformations into a static `{color: 'green'}` styles object so there'd be no runtime overhead of composition. I need to put some more thought into this.

dsslimshaddy:
Any update on this ?

rtsao:
@dsslimshaddy A HOC like this for Styletron already exists: https://github.com/tabazevedo/styletron-connect

rtsao:
Closing in favor of https://github.com/tabazevedo/styletron-connect

PR #133, opened 2017-05-22
Support for fallback values w/ guaranteed order

### LINKED ISSUE #6: Fully support value fallbacks
Value fallbacks (i.e. multiple values for a single property) are useful.

Because order is not currently guaranteed, using value fallbacks with styletron will not be robust.

faddee:
This looks hacky... and it is! :-) But I believe it's the simplest way to support fallback values without having to make major changes in the core and affect performance. Fixes #6 

Don't worry. I won't be sad if you decline this. :-P But it clearly works.

### DIFF
--- a/packages/styletron-server/src/cache-to-stylesheets-old-ie.js
+++ b/packages/styletron-server/src/cache-to-stylesheets-old-ie.js
@@ -20,7 +20,9 @@ function cacheToStylesheetsOldIE(cacheObj) {
       mediaSheets = getMediaSheets(cacheObj[key]);
       continue;
     }
-    if (typeof cacheObj[key] === 'string') {
+    if (typeof cacheObj[key] === 'object') {
+      ruleCount += Object.keys(cacheObj[key]).length;
+    } else {
       ruleCount++;
     }
     mainCss += baseHandler(key, cacheObj[key]);
@@ -44,7 +46,9 @@ function getMediaSheets(mediaObj) {
     let mediaCss = '';
     let ruleCount = 0;
     for (const key in obj) {
-      if (typeof obj[key] === 'string') {
+      if (typeof obj[key] === 'object') {
+        ruleCount += Object.keys(obj[key]).length;
+      } else {
         ruleCount++;
       }
       mediaCss += baseHandler(key, obj[key]);
--- a/packages/styletron-server/src/cache-to-stylesheets-old-ie.js
+++ b/packages/styletron-server/src/cache-to-stylesheets-old-ie.js
@@ -20,7 +20,9 @@ function cacheToStylesheetsOldIE(cacheObj) {
       mediaSheets = getMediaSheets(cacheObj[key]);
       continue;
     }
-    ruleCount += Object.keys(cacheObj[key]).length;
+    if (typeof cacheObj[key] === 'string') {
+      ruleCount++;
+    }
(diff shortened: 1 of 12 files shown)


rtsao:
I don't love the idea of the property names in the value string, but I think this actually pretty robust. I hadn't thought of this approach before, cool stuff! 👍 

There's a bit of a tradeoff here in terms of reduced de-duplication, but since this change fully guarantees order I think it's more robust implementation than what exists currently. Also, in practice, it seems that value fallbacks are often merely vendor prefixes, so this would have the benefit of collapsing these into a single atomic class (further compressing the CSS) sort of like https://github.com/rtsao/styletron/issues/104

Since this is related to https://github.com/rtsao/styletron/issues/104 I'd want to make sure this wouldn't cause any problems down the line.

I'm also open to updating the core to facilitating this in a first-class fashion, since admittedly this is pretty hacky and is essentially an abuse of the interface and implementation details.

What about something like:

```js
styletron.injectRawDeclaration(
  'color:red',
  ':hover',
  '@media (max-width: 400px)'
);

styletron.injectRawDeclaration(
  'color:red;color:blue',
  ':hover',
  '@media (max-width: 400px)'
);
```

The internal cache structure would need to be changed from:
```js
{
  color: {red: 'a'},
  pseudos: {
    ':hover': {
      color: {red: 'b'}
    }
  }
}
```
to:

```js
{
  'color:red': 'a',
  pseudos: {
    ':hover': {
      'color:red': 'b'
    }
  }
}
```

I'm not sure what the performance implications would be, but in my original benchmarking the former with the original `injectDeclaration` was faster because there's never any string concatenation performed.

My intuition is that this new cache format would be faster in the case of new styles, but slower in the case of existing styles. Probably worth doing some benchmarking.

faddee:
> I don't love the idea of the property names in the value string

LOL, you're funny. You surly would be high if you'd love it. ;-)

Sure, an `injectRawDeclaration` is a good idea. But do you mean it should complete the original `injectDeclaration` with a separate cache or replace it?

rtsao:
> But do you mean it should complete the original injectDeclaration with a separate cache or replace it?

I think the new cache format could just replace the original, then both `injectDeclaration` and `injectRawDeclaration` could shared the same cache. Additionally, I think at this point it might be worth deprecating the `injectDeclaration` method, then removing it from the next major version. Nobody really interacts with this low-level API directly so I don't see much value in having two different methods -- and since `injectRawDeclaration` is lower-level, we might as well support that one only.

faddee:
I've added `injectRawDeclaration` to:
- `styletron-core`
- `styletron-server`
- `styletron-client`

And `injectStyle` now uses `injectRawDeclaration` instead with full support of fallback values.

packages/styletron-server/src/cache-to-stylesheets-old-ie.js
```diff
@@ -20,7 +20,9 @@ function cacheToStylesheetsOldIE(cacheObj) {
       mediaSheets = getMediaSheets(cacheObj[key]);
       continue;
     }
-    ruleCount += Object.keys(cacheObj[key]).length;
+    if (typeof cacheObj[key] === 'string') {
```
faddee: I realized this count excludes pseudos. It needs to count keys if `cacheObj` is an object.

faddee:
Merged with #145

PR #134, opened 2017-05-23
Update test fixtures versions

rtsao:
Fixes https://github.com/rtsao/styletron/issues/131

Maybe?

### DIFF
--- a/packages/benchmarks/yarn.lock
+++ b/packages/benchmarks/yarn.lock
@@ -1267,7 +1267,7 @@ lexical-scope@^1.2.0:
   dependencies:
     astw "^2.0.0"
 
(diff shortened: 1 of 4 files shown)


faddee:
Sorry, it didn't help. Version 1.0.0 is the only version available. I get a similar error and the same info using Npm. It pretty clear the `test-fixtures` needs to be replaced.

image

rtsao:
Ah, I see. Have you tried `npm run bootstrap` instead? Lerna should link the local monorepo `test-fixtures` package instead of installing the public one. I think `yarn install`ing the monorepo packages directly won't symlink things appropriately.

faddee:
Oh, my bad. Sorry.

PR #145, opened 2017-06-06
Collapse prefixed styles into one atomic class

### LINKED ISSUE #104: [RFC] Collapse vendor prefixes into single atomic class
### What needs to be done?

Vendor prefixing needs to be rolled in to the core module, because it changes the implementation details of `injectDeclation`. It's possible this could be implemented as an alternate core module, so folks who don't want this don't have to use it.

### What are the benefits?

**Fewer classes**
- Nicer style inspection experience w/ DevTools out of the box
- Better theoretical performance
  - Fewer classes per element involved in style calculation
  - Shorter class strings (Fewer bytes over the wire)

**Possibly more robust SSR in the case of not shipping a static prefix map to the client.**
This is ~1.5 KB *gzipped* which is rather large in comparison to a Preact/lightweight app.

How does rolling up vendor prefixes into a single atom class make SSR in this case more robust? We can figure out at runtime which prefixes are required (for both values and properties). We can cache/memoize this too so there's only a startup cost. If we don't care about SSR, we can always do this. But if we care about SSR, ensuring our client-rendered markup is identical to the server-rendered markup can be difficult. We can guess based on user agent which prefixes are needed, but my thinking is this will never be 100% robust, and could lead to instances where extra/fewer atomic classes are produced in the client render than the server, leading to different markup. If vendor prefixes are rolled up into a single atomic class, then we can hydrate based on the original unprefixed declaration and ensure we never have different markup than the server render, as regardless of the resultant vendor prefixing, the same, single atomic class will be produced.

faddee:
This PR is on top of #133 and fixes #104.

In this implementation of `injectStylePrefixed` each entity is run through `prefixAll` by `inline-style-prefixer` and injects as a single atomic class using the `injectRawDeclaration`.

This will significantly reduce the number of class names which will theoretically have better performance, ease debugging and result in fewer bytes using SSR, see #104.

The computed block for each unique prefixed entity (property + value) is cached so it will only go through the prefixer once and should make e.g. renders in React faster. It only cache when the style value is a string or number since it's easier to cache primitives. I'm not sure whether complexity of caching arrays (fallback values) will be worth it since they are not that widely used.

### DIFF
--- a/packages/styletron-client/src/__tests__/browser.js
+++ b/packages/styletron-client/src/__tests__/browser.js
@@ -34,9 +34,8 @@ test('hydration basic', t => {
   const instance = new StyletronTest(elements);
   t.deepEqual(instance.getCache(), fixtures.basic.cache, 'cache hydrated');
   t.equal(instance.getUniqueDeclarationCount(), 5, 'count correctly hyrdated');
-  const newClass = instance.injectDeclaration({
-    prop: 'color',
-    val: 'purple',
+  const newClass = instance.injectRawDeclaration({
+    block: 'color:purple',
     media: '(max-width: 800px)',
   });
   t.equal(newClass, 'f', 'new class with correct count');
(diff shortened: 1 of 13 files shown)


rtsao:
Awesome work, this is great 🎉 !

> I'm not sure whether complexity of caching arrays (fallback values) will be worth it since they are not that widely used.

Agreed, probably not worth the complexity.

PR #158, opened 2017-07-10
Add `benchmark` dependency to packages/benchmark/package.json

steveluscher:
Unless _I'm_ missing something, this seems to have been missing.

### DIFF
--- a/packages/benchmarks/package.json
+++ b/packages/benchmarks/package.json
@@ -14,6 +14,7 @@
   "license": "MIT",
   "dependencies": {
     "aphrodite": "^1.0.0",
+    "benchmark": "^2.1.4",
     "browserify": "^13.1.0",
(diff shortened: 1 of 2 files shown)


rtsao:
The monorepo is set up to use Lerna's [hoist option](https://github.com/lerna/lerna/blob/master/doc/hoist.md#lerna-hoisting), and since `benchmark` is specified in the root package.json, I think everything ought to be okay as is.

steveluscher:
I guess that's what I was missing! All I know is that I went to the `benchmarks/` folder, ran `yarn install && npm start` and nothing worked. I've never used Lerna so I guess I needed a README.

¯\\_(ツ)_/¯

rtsao:
Yeah, I need create a CONTRIBUTING.md so it's clear how to get started.

PR #160, opened 2017-08-03
Enable a `selectorPrefix` on styletron-client

zanzamar:
Allows a "selectorPrefix" to be injected into the page for applications where a react app may be nested on a third party page.

Ensures that "resets" and other things can be applied and not "override" the attribute on the element.

Precludes the need to define every "reset" or "host page" override for every element in your code.

Utilized:
```
const styletron = new Styletron(styleElements, { selectorPrefix: '.SOME_WIDGET' });
```

Will continue to produce class names on the elements with their standard minification.  Class names injected into the `<style />` tags will have a prefix on them.    ie:

```
<style type="text/css">
.SOME_WIDGET b { color: 'red' }
</style>
```

### DIFF
--- a/packages/styletron-client/src/styletron-client.js
+++ b/packages/styletron-client/src/styletron-client.js
@@ -18,6 +18,7 @@ class StyletronClient extends StyletronCore {
    */
   constructor(serverStyles, opts) {
     super(opts);
+    this.selectorPrefix = opts ? opts.selectorPrefix : undefined;
     this.uniqueCount = 0;
     this.mediaSheets = {};
     if (serverStyles) {
@@ -78,7 +79,7 @@ class StyletronClient extends StyletronCore {
     const oldCount = this.uniqueCount;
     const className = super.injectDeclaration(decl);
     if (oldCount !== this.uniqueCount) {
-      const rule = declarationToRule(className, decl);
+      const rule = declarationToRule(className, decl, this.selectorPrefix);
       let sheet;
       if (decl.media) {
         if (!this.mediaSheets[decl.media]) {
@@ -103,9 +104,9 @@ module.exports = StyletronClient;
  * Injection helpers
  */
 
-function declarationToRule(className, {prop, val, pseudo}) {
+function declarationToRule(className, {prop, val, pseudo}, selectorPrefix) {
   const decl = `${prop}:${val}`;
-  let selector = `.${className}`;
+  let selector = `${selectorPrefix ? `${selectorPrefix} ` : ''}.${className}`;
   if (pseudo) {
     selector += pseudo;
   }


zanzamar:
Please share any thoughts you may have.  Pretty minimal change.  I could see this easily being a `2.5.10` branch.

rtsao:
Interesting idea. The question I have is what would be the use case difference between:

```js
const styletron = new Styletron(styleElements, { selectorPrefix: '.abc' });
```

```js
const styletron = new Styletron(styleElements, { prefix: '_abc' });
```

I think in both cases, there wouldn't really be any possibility of causing side effects (i.e. class collisions) on the host page. Aren't these basically equivalent?

Perhaps I don't quite understand, could you go into more detail in terms of what you mean by this:
> Ensures that "resets" and other things can be applied and not "override" the attribute on the element.
> 
> Precludes the need to define every "reset" or "host page" override for every element in your code.

zanzamar:
Hey Ryan.  Thank you for the thoughts.

Here is our scenario and why this is important.... It has to do with resets.

From reading the history of this project, it seems like your opinion is the resets should all be "applied" to every element.  Defining them "every time".

In our scenario we would like to apply reset styles and use "some" cascading as global rules...

On a single page we own.   Easy.  Global reset.

This becomes complicated when we are embedding our app on a third party site.  We want to have "reasonable resets"..  Without impacting the parent (imagine someone dropping in Stripe or Disqus onto their page..  Can't modify underlying styles).

So our reset file looks like this:

```
// reset.global.scss
.WIDGET_WRAP {
	div,
	span,
	... {
		margin: 0;
		padding: 0;
		border: 0;
		outline: 0;
		font-size: 100%;
		vertical-align: baseline;
		background: transparent;
		-webkit-text-size-adjust: 100%;
		position: relative;
	}

	... more resets...
}
```

If we do that, from a specificity standpoint then the individual Styletron classes on our elements will be "overridden".

If styletron generates class `b` with the attribute `padding: 10px` this will be overridden because `.WIDGET_WRAP div` has more specificity.

To overcome this, when writing the styles to the document we "prefix" them so that styletron generates:
```
<style type="text/css">
.WIDGET_WRAP b { padding: 10px }
</style>
```

This enables us have still a div looking like this:
`<div class="b">My Div</div>`

And the specificity of that div: `.WIDGET_WRAP .b` is "greater" then `.WIDGET_WRAP div`

With the P/R  this is as easy as:

`const styletron = new Styletron(styleElements, { selectorPrefix: '.WIDGET_WRAP' });`

Boom!

Please share your thoughts.

Example in action:
image

zanzamar:
FYI.  We ended up using this a few days ago as we wait for your thoughts:
https://gist.github.com/zanzamar/6b8cfd292346fb7237500238c1828a47

PR #173, opened 2017-10-02
Typed css properties

frenic:
Replaces `React.CSSProperties` with `csstype`. These covers CSS3 psuedo styles like
```ts
{
  color: 'white',
  '::before': {
    display: 'inline-block',
  }
}
```

The problem is still custom `@media` style where you'll have to explicitly type it until TypeScript has solved the misinterpretation of key value types:

```ts
import {StyleType} from 'styletron-react';

const Pretty = styled(div, () => {
  const media: StyleType = {
    color: 'purple',
  };
  return {
    color: 'pink',
    '@media (min-width: 768px)': media,
  }
});
```

Added a type for the `styleProps`-property of `styled` but there's no way to type additions to that property when composing until TypeScript [rest types](https://github.com/Microsoft/TypeScript/pull/13470) has been released.

### DIFF
--- a/packages/styletron-react/index.d.ts
+++ b/packages/styletron-react/index.d.ts
@@ -35,6 +35,8 @@ declare namespace StyletronReact {
 
   type ClassNameProp = {className?: string};
 
+  type StyleProp<TProps> = {styleProps: TProps};
+
   export function core<TOuterProps, TInnerProps>(
     base: React.StatelessComponent<TInnerProps>,
     style: StyleArgument<TOuterProps>,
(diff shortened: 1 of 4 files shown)


pspeter3:
Is there any reason not to merge this?

faddee:
No, maybe there's some confusion regarding `@media` in my example.

```js
{
    color: 'pink',
    '@media (min-width: 768px)': {
      color: 'purple',
    },
}
```
This will still work but `{ color: 'purple' }` in this case will by typed as `any`. So you need to pass the `@media` style objects through a function (like `typedStyle(style: StyleType) => style`) or define it as a typed variable (like in the example) to enable type checking.

rtsao:
Per https://github.com/styletron/styletron/issues/228 I want to use the normal csstype dependency, but still need to work through issues w/ Flow. Closing this for now. Thanks!

PR #186, opened 2017-11-30
Fix warning about misspelled propTypes

jbellsey:
React 16.1 now warns that StyletronProvider is trying to set "PropTypes" instead of "propTypes" (note capital).


rtsao:
Nice catch, thanks!

PR #191, opened 2017-12-14
Fix incorrect default import in styletron-preact

crohlfs:
I'm running into issues with the preact bindings trying to import from a default export that doesn't exist in `styletron-utils`. I'm using webpack.

I noticed that the inferno and react bindings use this import instead, using this resolves the issue.

Let me know if any tests or anything are required.


rtsao:
Thanks for the fix!

PR #194, opened 2018-01-01
Avoid rendering empty class attribute

ratson:
When `className` is blank, `class=""` should not be rendered.


rtsao:
Nice! Appreciate the PR 😄

Issue #161, opened 2017-08-04
Style inheritance not allowing overrides.


Shent:
I've started styling my Core components and everything was perfect until I started to Extend them.
The extended component inherits all the core styles BUT with a priority, so I am not able to override them. This is a fundamental issue and a total blocker when you have a component composition and inheritance which we have in any React project nowadays.

Broken example is provided here: https://codesandbox.io/s/k5oLQ4YZY 
My expectation is the MenuIcon to be a lot bigger, nothing happens.
If I move the ```<Icon />``` component 1 line above, so it is rendered before the ```<MenuIcon />``` - it works but I do not want the Icon at all on this page - that makes it break.

If I implement the same example with only styled-components it works, so the issue comes from Styletron implementation for sure.

Shent:
I have a suspicion that the way you parse the styles could be wrong if you are traversing through all the components in the order they are declared. Then the issue will be present for sure because a Core component will be seen after it's derived components and the core styles will be defined after the ones you want to prioritize. So this would not be easy to fix though. All the parsing logic needs to be reworked I guess.

rtsao:
Hey Shent,

From your example, you have this:
```js
const GenericIcon = styled('div', {
  width: '1em',
  height: '1em',
  border: '1px solid blue',
  color: 'red'
})

class Icon extends Component {
  render() {
    return (
    	<GenericIcon className={this.props.className}>Icon</GenericIcon>
    )
  }
}

const MenuIcon = styled(Icon, {
  width: '150px',
  height: '150px'
});

```

For composition of styled elements, you need to pass in *another styled element*. So you would need to do something like this instead:

```js
const GenericIcon = styled('div', {
  width: '1em',
  height: '1em',
  border: '1px solid blue',
  color: 'red'
})

const MenuIcon = styled(GenericIcon, {
  width: '150px',
  height: '150px'
});

```

Shent:
Hello Ryan -

I think you did not the idea of the issue I was having there.
If I am using Pure Functions as Input it is all fine, I got that. 
The problem is I have no pure functions in my project due to its specificity but Components that have functions I need to extend.
As in the example, I have a GenericIcon that has certain functions and I have to extend from it to make MenuIcon, SocialIcon, FooterIcon etc. 

Is there any way to accomplish this with Styletron or I should simply stick with StyledComponents because the concept works there, although the outputted CSS will be 30 times bigger?

rtsao:
[...]

Issue #174, opened 2017-10-11
Composable styled api


ganemone:
Looking for some feedback on a small change to the `styled` api that could potentially allow for some interesting avenues for composition.

Looking at an example in the documentation:

```jsx
import {styled} from 'styletron-react';

const Panel = styled('div', (props) => ({
  backgroundColor: props.alert ? 'orange' : 'lightblue',
  fontSize: '12px'
}));

const DeluxePanel = styled(Panel, (props) => ({
  backgroundColor: props.alert ? 'firebrick' : 'rebeccapurple',
  color: 'white',
  boxShadow: '3px 3px 3px darkgray'
}));
```

If we changed the `styled` interface from: 

`styled(base, styleFn) → {function}` to
`styled(styleFn) → (base) → {function}`

We could rewrite the above example as:

```jsx
import {styled} from 'styletron-react';

const SimpleStyles = styled((props) => ({
  backgroundColor: props.alert ? 'orange' : 'lightblue',
  fontSize: '12px'
}));

const FancyStyles = styled((props) => ({
  backgroundColor: props.alert ? 'firebrick' : 'rebeccapurple',
  color: 'white',
  boxShadow: '3px 3px 3px darkgray'
}));

const Panel = SimpleStyles('div');
const DeluxePanel = compose(SimpleStyles, FancyStyles)('div');
```

You could further break this down into something like this (somewhat contrive) example:

```jsx
import {styled} from 'styletron-react';

const Small = styled({
  fontSize: '12px',
});

const SimpleAlert = styled((props) => ({
  backgroundColor: props.alert ? 'orange' : 'lightblue',
}));

const Shadow = styled({
  boxShadow: '3px 3px 3px darkgray'
});

const FancyAlert = styled((props) => ({
  backgroundColor: props.alert ? 'firebrick' : 'rebeccapurple',
  color: 'white',
}));

const Panel = compose(Small, SimpleAlert)('div');
const DeluxePanel = compose(Small, Shadow, FancyAlert)('div');
```

It is a minor change, but potentially encourages designing your styled components in a way where you compose base styles throughout your application. There are also several variations on this API that could achieve the same results. Such as `import {styledDiv}` or `styled('div')(styleFn)`, etc. The important thing is encouraging unary functions for composability.

Thoughts? Not sure if this is a good pattern or not.

bzalasky:
@ganemone Personally, I like this approach. You can enable this kind of API with a small higher-order component using Recompose.

```javascript
[...]

PR #214, opened 2018-01-19
styletron-react: better composition

josepot:
Hi and thanks for this project... I love it!

So this PR attempts to solve 2 different problems that you probably didn't know that `styletron-react` had... 😄  Lets see if I can convince you 😬 

Lets start with the "silly/easy" one, let me explain it with one example:

```jsx
import {styled} from 'styletron-react';

const Panel = styled('div', {
  ':hover': {
    fontWeight: 'bold',
    color: 'blue',
  },
  fontSize: '12px'
});

const DeluxPanel = styled(Panel, {
  ':hover': {
    color: 'red',
  }
)
```

It would be nice if `DeluxPanel` on hover was red **and** bold. With the current implementation it will be only "red" (and not "bold"). Same issue happened with media-queries, basically that the objects were being replaced rather than merged.

Now lets have a look at the second improvement. This one is a bit tougher to explain, but it fixes issue #161 and IMHO it will be 100% necessary if we ever go down the path of #174. Again, let me explain it with some examples.

Lets say that I have a Currency component: that looks like this:

```jsx
import {connect} from 'react-redux';
import {injectIntl} from 'react-intl';
import {styled} from 'styletron-react';
import {compose, mapProps, setDisplayName} from 'recompose';

const enhancer = compose(
  setDisplayName('Currency'),
  injectIntl,
  connect(({currency}) => ({currency}), {}),
  mapProps(({currency, intl, children, ...rest}) => ({
    children: intl.formatNumber(children, {style: 'currency', currency}),
    ...rest,
  })
);

const Currency = enhancer(styled('span', {
  color: 'blue',
  textSize: '1em',
}));

Currency.propTypes = {
  children: PropTypes.number.isRequired,
};

export default Currency;
```

I can use my "awesome" `Currency` component like this `<Currency>5</Currency>` and it will display the number formatted with the currency that the user has selected.

Lets say that somewhere else I need to have a `BigCurrency` component, and I don't want to add another prop to the `Currency` component (just for the sake of argument). Basically, I would like to be able to do something like this:

```jsx
const BigCurrency = styled(Currency, {textSize: '2em'});
```

However, with the current implementation that won't work: because the `span` will receive both classes: the class of `textSize: 1em` and the class of `textSize: 2em`. So, what class will take over? It's a lottery.

If this PR gets merged, then `BigCurrency` will generate a `span` that will only receive the class of `textSize: 2em` and not the other, which I think it's awesome!

I'm well aware that I wouldn't have had this issue if I had implemented the `Currency` component like this:

```jsx
import {connect} from 'react-redux';
import {injectIntl} from 'react-intl';
import {styled} from 'styletron-react';
import {compose, mapProps, setDisplayName} from 'recompose';

const CurrencySpan = compose(
  setDisplayName('Currency'),
  connect(({currency}) => ({currency}), {}),
  injectIntl,
  mapProps(({currency, intl, children, ...rest}) => ({
    children: intl.formatNumber(children, {style: 'currency', currency}),
    ...rest,
  })
)('span');

const Currency = styled(CurrencySpan, {
  color: 'blue',
  textSize: '1em',
}));

Currency.propTypes = {
  children: PropTypes.number.isRequired,
};

export default Currency;
```

So, maybe this is not the best example, but before I jump into a -perhaps- more compelling example. I do want to point out that it's a pity that `styled` doesn't behave well with "normal" HOC composition.

Let me try with a hopefully more compelling example:

```jsx
import React from 'react';
import {styled} from 'styletron-react';

const Wrapper = styled('div', {
  border: '1px solid black',
  padding: '0.3 em',
});

const Title = styled('div', {
  fontSize: '1.2 em',
  color: 'grey',
});

const TitleBox = ({title, children, ...props}) => (
  <Wrapper {...props} >
    <Title>{title}</Title>
    {children}
  </Wrapper>
);

export default TitleBox;
```

Wouldn't it be nice to be able to do this?

```jsx
const DashedTitleBox = styled(TitleBox, {
  border: '1px dashed black',
});
```

With the current implementation `DashedTitleBox` will receive both "border" classes... It would be nice if it only received the "dashed" one.

## Full disclosure:

The problem that I have is that I don't used `styled` directly, I use my "own" version that looks (more or less) like this:

```js
import {styled} from 'styletron-react'
export default fnOrObj => Component => styled(Component, fnOrObj);
```

Because to me the styletron `styled` fn is a HOC (it's a function that takes a component as an argument and returns an enhanced one), and since I want to be able to compose it like the rest of HOCs, I've "flipped" and "curried" it...  However, as you can imagine, with the current implementation that makes my life quite miserable.

### DIFF
--- a/packages/styletron-client/src/index.js
+++ b/packages/styletron-client/src/index.js
@@ -60,7 +60,8 @@ class StyletronClient extends StyletronCore {
           pseudo: decl[2],
           media,
         },
-        decl[1]
+        decl[1],
+        this.declarations
       );
     }
   }
--- a/packages/styletron-core/src/__tests__/index.js
+++ b/packages/styletron-core/src/__tests__/index.js
@@ -22,6 +22,12 @@ test('test injection', t => {
   instance.injectDeclaration(decl1);
   t.equal(instance.getCache()['color:red'], 'a');
   t.equal(instance.getCachedDeclaration({block: 'color:red'}), 'a');
+  let media, pseudo;
+  t.deepEqual(instance.getDeclarationFromClassName('a'), {
+    block: 'color:red',
+    media,
+    pseudo,
+  });
   t.equal(instance.getCount(), 1, 'unique count incremented');
   instance.injectDeclaration(decl1);
   t.equal(
@@ -115,6 +121,7 @@ test('test injection with prefix', t => {
   instance.injectRawDeclaration(decl1);
   t.equal(instance.getCache()[block1], 'qqa');
   t.equal(instance.getCachedDeclaration(decl1), 'qqa');
+  t.deepEqual(instance.getDeclarationFromClassName('qqa'), decl1);
   t.equal(instance.getCount(), 1, 'unique count incremented');
   t.end();
 });
--- a/packages/styletron-core/src/index.js
+++ b/packages/styletron-core/src/index.js
@@ -13,6 +13,7 @@ class StyletronCore {
       media: {},
       pseudo: {},
     };
+    this.declarations = {};
     this.prefix = prefix === '' ? false : prefix;
     this.uniqueCount = 0;
     this.offset = 10; // skip 0-9
@@ -20,7 +21,7 @@ class StyletronCore {
     this.power = 1;
   }
 
-  static assignDecl(target, decl, className) {
+  static assignDecl(target, decl, className, declarations) {
     const {block, media, pseudo} = decl;
     let targetEntry;
     if (media) {
@@ -38,6 +39,7 @@ class StyletronCore {
       targetEntry = targetEntry.pseudo[pseudo];
     }
     targetEntry[block] = className;
+    declarations[className] = decl;
   }
 
   /**
@@ -69,7 +71,7 @@ class StyletronCore {
     const virtualCount = this.incrementVirtualCount();
     const hash = virtualCount.toString(36);
     const className = this.prefix ? this.prefix + hash : hash;
-    StyletronCore.assignDecl(this.cache, decl, className);
+    StyletronCore.assignDecl(this.cache, decl, className, this.declarations);
     return className;
   }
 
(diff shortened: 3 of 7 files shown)


rtsao:
Thanks for the nice write up! 😄 I agree composition with wrappers is problematic with the current API.

I'm working on a LTR-to-RTL feature and was dealing with a similar issue.

I came up with a somewhat different solution (as part of the proposed Styletron v4 API: https://github.com/rtsao/styletron/issues/215, see the bottom) which I think should be robust and easy to work with. Let me know what you think!

Some background: I want to apply a RTL transform depending on a global context, but context can't be updated in React (some sort of pub/sub that triggers a state change is needed). I don't want this functionality directly in the core styled elements, so I need a wrapper on each styled element. But I want composition to still work, even with the wrappers.

josepot:
Hi @rtsao and thanks for getting back to me so quickly!

First of let me admit that the implementation that I'm suggesting looks a bit hacky... I have an idea to make it more robust, but I haven't had time to implement it yet.

That being said, I think that this changes are compatible with v3, in the sense that they are 100% backwards compatible... As long as we consider that the "issue" when composing pseudo and media queries was a bug and not a feature... Now I'm starting to realize that I should have broken this PR in 2 😅 In any event, I think that it won't hurt to add this to v3, but it's your library so I totally get it if you don't want to merge this until we find a more robust solution.

About the new API that you are suggesting for v4... I can't help to wonder: wouldn't it be a lot easier to change the current API to something like: `curry(flip(currentStyled))` and encourage for composition solve all those problems?

Then the issue of whether composing from RTL or LTR wouldn't need to be set by context, it's just a matter of using `compose` or `pipe`.

I like the `$as` built-in prop that you are suggesting for v4, but I think that I wouldn't use it that often because most of the time I would have my "style-enhancers" that I would then use with whichever "base" I want.

rtsao:
I think this should be possible with `withTransform`, etc. Closing this since the codebase has changed dramatically since this PR was opened. Definitely open to adding additional composition functions provided typing can work properly. Thanks!

PR #222, opened 2018-04-04
Syntax mistake in README

kahwee:
I think you made a mistake in your README. 

Changed:

```
  return {backgroundColor: props.$alert < "orange" : "lightblue"};
```

to:

```
  return {backgroundColor: props.$alert ? "orange" : "lightblue"};
```


rtsao:
Nice catch, thanks!

PR #231, opened 2018-05-01
Debug classes w/ source map to JS location

rtsao:
In development, add a unique debug class name to elements rendered by a given styled component. This debug class name has no styles but a generated sourcemap to JS location. This sourcemap is generated dynamically by capturing a stack trace, no build-time plugins are needed.

This makes it easier to map from a DOM element to the styled component using the DOM inspector.

**Demo**


frenic:
Nice work!

PR #251, opened 2018-06-11
Use css-to-js-sourcemap-worker

rtsao:
- Replaces built-in sourcemapping


c58:
It have a mistake in docs: It is saying `DebugUtility` but it is actually `DebugEngine` @rtsao

c58:
Also please do not miss this issue, the worker is not working for me: https://github.com/rtsao/styletron/issues/254

PR #257, opened 2018-07-03
Fix withStyleDeep docs

schnerd:
Docs seemed like they could be improved here a bit.


schnerd:
[styletron-react-core](https://github.com/rtsao/styletron/tree/master/packages/styletron-react-core) docs also have links for withStyle, withStyleDeep, withTransform, withWrapper, that do nothing. Happy to copy docs into that README as well or link to function docs in `styletron-react` README

rtsao:
Awesome, thanks!

PR #274, opened 2018-10-02
Skip object prototypes when serializing keyframes

ksheedlo:
This change fixes a bug in `keyframesToBlock` that was causing properties on the prototype to be serialized in a tag.

### DIFF
--- a/packages/styletron-engine-atomic/src/css.js
+++ b/packages/styletron-engine-atomic/src/css.js
@@ -1,5 +1,7 @@
 // @flow strict
 
+declare var __DEV__: boolean;
+
 import hyphenate from "./hyphenate-style-name.js";
 
 export function atomicSelector(id: string, pseudo: string): string {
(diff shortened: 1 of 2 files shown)


packages/styletron-engine-atomic/src/css.js
```diff
@@ -11,9 +13,20 @@ export function atomicSelector(id: string, pseudo: string): string {
 }
 
 export function keyframesToBlock(keyframes: Object): string {
+  if (__DEV__ && typeof Object.getPrototypeOf(keyframes) !== "undefined") {
+    if (Object.getPrototypeOf(keyframes) !== Object.getPrototypeOf({})) {
+      // eslint-disable-next-line no-console
+      console.warn(
+        "Only plain objects should be used as animation values. Unexpectedly recieved:",
```
ksheedlo: nit: `received`

ksheedlo:
@rtsao I feel kind of ambivalent as to whether we should warn about this or not. I found out the most likely reason this issue affected my application is by loading https://github.com/Marak/colors.js/blob/v0.6.2/colors.js on the server. Do you think it's worth it to warn about prototypes if we're just going to skip over them?

Anyway your edits LGTM 👍

gergelyke:
closing in favor of https://github.com/styletron/styletron/pull/300

Issue #277, opened 2018-10-30
Missing flow property


ardok:
https://github.com/styletron/styletron/blob/master/packages/styletron-standard/src/style-types.js#L1218

https://developer.mozilla.org/en-US/docs/Web/CSS/justify-content
```
justify-content: start;      /* Pack items from the start */
justify-content: end;        /* Pack items from the end */
justify-content: left;       /* Pack items from the left */
justify-content: right;      /* Pack items from the right */
justify-content: stretch; 
justify-content: safe center;
justify-content: unsafe center;
```

frenic:
The types should be updated with the latest from https://github.com/frenic/csstype/blob/master/index.js.flow. Probably more properties missing since it was last updated.

Issue #228, opened 2018-04-18
Use csstype as a dependency 


frenic:
A lot has happen since you copyed it. Bugs has been fixed and a lot of properties has been added. And more features and updates are coming. Using it as a dependency would be beneficial instead of updating the types by hand.

If it doesn't work, I would like to know why?

Another thought: In TypeScript-world libraries sees csstype as an ability to provide closed typings for style. I know it's not default in Flow. But you're able to do it using the `$Exact` and `$Shape` utility. I can't see whether you already use this or not. Hade a look in the [type check file at csstype](https://github.com/frenic/csstype/blob/master/typecheck.js) for examples. But maybe this isn't the way to go for Flow users?

rtsao:
Yeah, I'm working on migrating over; I definitely want to get rid of all the vendored code.

The times when I've tried switching over, for some reason Flow seems to hang -- I'm still not entirely sure what's going on, but I suspect it might have to do with just how many types there were in the exported file. When I deleted all the unused type definitions (e.g. hyphenated properties, etc), it seemed to work. It seems weird that this would make any difference, but I wonder if the sheer amount of types is overwhelming the flow server.

frenic:
Oh, that's weird. I mean, defining all the DOM APIs and that kind of stuff should be even more. Could it have anything to do with the file size? It's over 200KB+. I could provide csstype with a split version of the Flow types if performance is an issue. But I would like to be sure that it isn't a recursion issue or something like that first.

rtsao:
Just documenting here the reason csstype was vendored. Things may have changed with newer versions of Flow, however.

- Generics for length type seemed to cause breakages. Removing the generics (e.g. length type) resolved the problem.
- All the unused Flow types seemed to overwhelm Flow (it would basically just behave like `any`). Deleting the unused permutations resolved the issue.

We should upstream these potentially.

Issue #226, opened 2018-04-17
Missing SVG types


KevinGrandon:
> [flow] Cannot call `styled` with object literal bound to the second parameter because string [1] is incompatible with object type [2] in property `fill`. (References: [1] [2])

```js
const Svg = styled('svg', {
  fill: 'currentColor',
});
```

PR #279, opened 2018-11-28
Improve flow types, remove styletron-react-core

rtsao:
## Remove style object generic parameterization in Flow types

Flow 0.85 requires explicit generic annotations for exported styled components. Because of this, the existing parameterizing of style object types becomes rather burdensome. By removing this and hardcoding to a the standard style object shape, it won't be necessary to specify style object type in these annotations.

Removing this parameterization obviates the main purpose of the `styletron-react-core` package, so this PR moves the `createStyled` function from `syletron-react-core` into `styletron-react` and deletes `styletron-react-core`.

Additionally, I don't think this parameterization was ever used in practice, so removing this also makes sense for that reason.

## Use csstype

It seems the bugs in Flow that were preventing usage of `csstype` have been resolved. This fixes many issues with incorrect style object types.

- Fixes https://github.com/styletron/styletron/issues/226
- Fixes https://github.com/styletron/styletron/issues/228
- Fixes https://github.com/styletron/styletron/issues/277


packages/flow-type-tests/component-base.js
```diff
@@ -0,0 +1,59 @@
+// @flow
```
AlexMSmithCA: Is this file for manually spotchecking Flow?

rtsao: Yep, the flow is using the --max-warnings=0 flag so if there are unused suppressions this will cause an error as well (which can catch when something *should* error but does not).

PR #289, opened 2019-01-08
adds babel plugin to apply displayName to styletron-react components

chasestarr:
before
```js
const Foo = styled("div", {
  color: "red"
});
const Bar = styled("div", {
  color: "red"
});
```

after
```js
Foo = styled("div", {
  color: "red"
});
Foo.displayName = "Foo";
Bar = styled("div", {
  color: "red"
});
Bar.displayName = "Bar";
```

This automates a fairly tedious process of [manually applying the displayName](https://github.com/uber-web/baseui/pull/583/files#diff-42e8c15b743e67eede727008010041c5R20) to each styletron component. Based the plugin almost entirely on a [similar plugin](https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-react-display-name) for applying displayName on React.createClass calls.


rtsao:
This is awesome, thanks!

I think this could be made slightly more robust in a couple ways:

1. Checking against specific package names to avoid false positives result from name collisions, for example:

```js
import styled from "contrived-package";

const ContrivedExample = styled("https://bootstrap.com/min.css" {
  cache: false
});
```

2. Allowing for different local import specifiers to avoid false negatives, for example:
```js

import {styled as other} from "styletron-react";

const Foo = other("div", {color: "red"});
```

This could be implemented with something like this: https://astexplorer.net/#/gist/56b656602df1ec9882ae7136f5ee586e/d921259fcdbe5400611fe6004c2fb44d51061e9d

chasestarr:
Sounds good. I'll update to follow this pattern. The main problem that I want to solve is in the baseui project, where we modify the styletron `styled` function and export ourselves. Are you fine with me adding baseui to this list? `["styletron-react", "fusion-plugin-styletron-react"]`

rtsao:
Yeah, that makes sense!

PR #296, opened 2019-03-04
Sort media queries, mobile first 

tajo:
## Problem

Styletron currently doesn't do anything about the order of media queries and they can be injected in a random order as components get mounted. However, the order of media queries matters since:

```
(min-width: 480px)
(min-width: 600px)
```

and

```
(min-width: 600px)
(min-width: 480px)
```

would yield different results when the viewport's width is over 600px - the last matched definition wins. In mobile-first approach, it's preferred to enforce the ascendant order.

## Solution

`styletron-engine-atomic` can parse media queries and keep them sorted. There is a new variable `MutliCache.sortedCacheKeys` added for this purpose. When a new `media` key appears (cache miss), it's added to the end of `sortedCacheKeys` and the the `sortedCacheKeys` array is sorted using this [implementation](https://github.com/dutchenkoOleg/sort-css-media-queries) (without the `desktopFirst` part).

`StyletronServer` then uses `sortedCacheKeys` in `stringify` and `sheetify` methods to enforce this order correctly. `StyletronClient` doesn't need the whole `sortedCacheKeys` array but only the position where the new `media` stylesheet should be inserted through the `insertAtIndex` variable.

### Mobile first vs desktop first

You can't have it both so Styletron will do mobile-first by default.

### Perf
I didn't want to touch / rewrite the css sort so I didn't do one slight optimization: Since we are building a sorted `sortedCacheKeys` array from scratch we could use more efficient algorithm `O(n)` (or even `O(logn)`) where `n` is number of media queries. This implementation sorts `sortedCacheKeys` with every new key so it's `O(n*logn)`. However, typical application has only a few different breakpoints and this sort happens only when the cache is not hit, so it has almost zero impact. However, if it's important to you, I can spend more time and update it (although it would come with more code / bundle size).

// cc @gergelyke @chasestarr @nadiia

### DIFF
--- a/packages/styletron-engine-atomic/src/cache.js
+++ b/packages/styletron-engine-atomic/src/cache.js
@@ -28,9 +28,7 @@ export class MultiCache<T> {
       cache.key = key;
       this.sortedCacheKeys.push(key);
       this.sortedCacheKeys.sort(sortMq);
-      const insertAtIndex = this.sortedCacheKeys.findIndex(
-        media => media === key,
-      );
+      const insertAtIndex = this.sortedCacheKeys.indexOf(key);
       this.caches[key] = cache;
       this.onNewCache(key, cache, insertAtIndex);
     }
--- a/packages/styletron-engine-atomic/src/cache.js
+++ b/packages/styletron-engine-atomic/src/cache.js
@@ -1,12 +1,14 @@
 // @flow
 
 import SequentialIDGenerator from "./sequential-id-generator.js";
+import sortMq from "./sort-css-media-queries.js";
 
 export class MultiCache<T> {
   caches: {[string]: Cache<T>};
   idGenerator: SequentialIDGenerator;
-  onNewCache: (string, Cache<T>) => any;
+  onNewCache: (string, Cache<T>, number) => any;
   onNewValue: (cache: Cache<T>, id: string, value: T) => any;
+  sortedCacheKeys: string[];
 
   constructor(
     idGenerator: SequentialIDGenerator,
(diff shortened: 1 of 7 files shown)


packages/styletron-engine-atomic/src/cache.js
```diff
@@ -16,18 +18,28 @@ export class MultiCache<T> {
...
     this.caches = {};
   }
 
   getCache(key: string): Cache<T> {
     if (!this.caches[key]) {
       const cache = new Cache(this.idGenerator, this.onNewValue);
       cache.key = key;
+      this.sortedCacheKeys.push(key);
+      this.sortedCacheKeys.sort(sortMq);
+      const insertAtIndex = this.sortedCacheKeys.findIndex(
```
rtsao:
```suggestion
      const insertAtIndex = this.sortedCacheKeys.indexOf(key);
```

packages/styletron-engine-atomic/src/cache.js
```diff
@@ -16,18 +18,28 @@ export class MultiCache<T> {
...
+        media => media === key,
```
rtsao:
```suggestion
```

packages/styletron-engine-atomic/src/cache.js
```diff
@@ -16,18 +18,28 @@ export class MultiCache<T> {
...
+      );
```
rtsao:
```suggestion
```

packages/styletron-engine-atomic/src/cache.js
rtsao: I think `indexOf` might be better since it won't require any Array polyfills in older browsers. IIRC this codebase works with only prepublish transpilation.

rtsao:
Agreed the tradeoff for sorting is probably not worth the complexity or bundle size since there is rarely going to be more than a dozen or two media queries.

PR #300, opened 2019-03-21
fix(keyframes): skip object protos

gergelyke:


packages/styletron-engine-atomic/src/css.js
```diff
@@ -17,11 +17,22 @@ export function keyframesToBlock(keyframes: {[string]: Object}): string {
...
+        keyframes,
+      );
+    }
+  }
   let result = "";
   for (const animationState in keyframes) {
-    result += `${animationState}{${declarationsToBlock(
-      keyframes[animationState],
-    )}}`;
+    if (Object.prototype.hasOwnProperty.call(keyframes, animationState)) {
```
rtsao: I'm wondering whether this runtime check is required if we have development warnings. Maybe just the development warnings will suffice?

Everywhere else in styletron it is assumed everything is always a POJO.

gergelyke: that sounds good to me

Issue #284, opened 2018-12-13
styletron-react v5 discussion


rtsao:
### Reducing burden of Flow generic annotations

[The requirement of Flow 0.85+ to annotate generics is definitely a bit burdensome](https://github.com/styletron/styletron/blob/master/docs/flow.md). `styletron-react@4.4.0` makes these annotations much more palatable, but having to annotate generics at all is definitely an increase in friction.

I think it is worth looking at adding an alternative API that don't require these annotations (for example a `StyletronComponent`). Definitely some more thinking to be done here.

### Deep object merging by default

It has been brought up `withStyle` should be deep object merge because this is the common use case. I am inclined to agree with this, as specific transformations (such as replacements) should just use the more generic `withTransform`. One option is making this change (as a breaking change), but the other option is just creating new names (as part of https://github.com/styletron/styletron/issues/283) and deprecating the old ones so migration can be more gradual. Writing codemods for this should be fairly trivial in any case.

### Allowing mix of arbitrary HOCs in-between Styletron HOCs

This would allow for more flexibility of composition (i.e. allowing for `connect`, `withRouter`, etc. between `withStyle` HOCs)

### Hooks

One of the benefits of the styled-components pattern is the lack of footguns related to manual composition of atomic class strings.

React hooks has some pretty significant footguns, but relies on lint rules and convention to avoid them. We could take a similar approach with a hooks-based Styletron API by adding a lint rule to guard against manual class name composition.

tajo:
## Hooks

How about this?

```jsx
const Test = () => {
  const { css } = useStyletron();
  return (
    <div className={css({ color: red })}>Red</div>
  )
}
```

I agree that it adds some footguns but it also removes some hurdles:
- defining and naming styled components upfront
- `$prop` filtering
- impossible to combine Styletron's styles + 3rd party styles
- it's more readable, less typing? 

## Themes

It seems really useful and convenient to support themes out of the box. Also, it could be integrated into the hook:

```jsx
const Test = () => {
  const { css, theme } = useStyletron();
  return (
    <div className={css({ color: theme.colors.red })}>Red</div>
  )
}
```

Otherwise we would force users to use a second hook just to get the theme or creating a custom `useStyletron` hook.

## Current selector

```jsx
styled('div', {
  color: 'red',
  '&.nested': { color: 'blue' }
})
```

would output

```css
.a {
  color: red
}

.b.nested {
  color: blue
}
```

and 

```html
<div class="a b" />
```

[...]

Issue #256, opened 2018-07-02
No-op engine by default


rtsao:
In testing scenarios, it might be preferable to have a no-op engine by default (i.e. if a styletron engine is never provided via React Context).

If this were to be added, I think it would make sense to log warnings if `NODE_ENV !== 'test'` and any engine method were called, just so in development it is clear that an engine is required.

tajo:
Done.

PR #304, opened 2019-04-03
Add useStyletron hook and default no-op engine

tajo:
solves https://github.com/styletron/styletron/issues/256, https://github.com/styletron/styletron/issues/284#issuecomment-467194731

**EDIT: Descoped, explained [bellow](https://github.com/styletron/styletron/pull/304#issuecomment-479652251).**

# useStyletron

`useStyletron` is a React [hook](https://reactjs.org/docs/hooks-intro.html). 

This API introduces a very direct way of styling components (it's more discussed in https://github.com/styletron/styletron/issues/284#issuecomment-467194731). It should be handy for non-customizable components since it's pretty concise (no need to name the styled components) and prevents multiple pitfalls (ref forwarding, prop filtering...).

**It also supports theming**.

```jsx
import { Provider, useStyletron } from "styletron-react";

export default () => {
  const [ css ] = useStyletron();
  return (
    <div className={css({ color: 'blue' })}>Blue text</div>
  );
};
```

# No-op engine

If you forget to wrap your application with `<Provider>`, Styletron switches to a no-op mode and prints out a warning (instead of crashing). This is nice when writing tests. You don't need to setup Styletron if you don't care about styles. However, you should set `NODE_ENV = test` to prevent the warnings.

# Supporting only React 16.3+ / 16.8+

**This PR should be landed as a part of v5 milestone.** Why?

I believe it makes sense to drop the support for React v15 since v16 was released [18 months ago](https://github.com/facebook/react/releases/tag/v16.0.0). 

v16.3 introduces ref forwarding and createContext APIs. This PR removes `create-react-context` polyfill to save some bytes and we are going to implement ref forwarding in a few weeks as well. So 16.3+ is hard minimum.

If you want to use the `useStyletron` hook, you need to be on v16.8. Since it should be fairly easy to upgrade React from v16.3 to v16.8, I set the peerDependency to v16.8.

# Todos

- debugEngine is broken in the `useStyletron` API, I need to look more into how the whole thing works (I was not able to make the sourcemaps work yet even for the current version). Also, we could fix it later. 
- since this will be released as v5, docs will be updated later

### DIFF
--- a/packages/styletron-react/src/index.js
+++ b/packages/styletron-react/src/index.js
@@ -6,7 +6,6 @@ declare var __DEV__: boolean;
 declare var __BROWSER__: boolean;
 
 import * as React from "react";
-import createReactContext, {type Context} from "create-react-context";
 import {
   driver,
   getInitialStyle,
@@ -30,9 +29,9 @@ import {addDebugMetadata, DebugEngine} from "./dev-tool.js";
 
 export {DebugEngine};
 
-const StyletronContext: Context<any> = createReactContext();
-const HydrationContext = createReactContext(false);
-const DebugEngineContext = createReactContext();
+const StyletronContext = React.createContext<*>();
+const HydrationContext = React.createContext(false);
+const DebugEngineContext = React.createContext();
 
 type DevProviderProps = {
   children: React.Node,
--- a/packages/styletron-react/src/index.js
+++ b/packages/styletron-react/src/index.js
@@ -4,6 +4,7 @@
 
 declare var __DEV__: boolean;
 declare var __BROWSER__: boolean;
+declare var process: any;
 
 import * as React from "react";
 import {
@@ -33,6 +34,7 @@ const StyletronContext = React.createContext<StandardEngine>({
   renderStyle: () => "",
   renderKeyframes: () => "",
   renderFontFace: () => "",
+  noopEngine: true,
 });
 const HydrationContext = React.createContext(false);
 const DebugEngineContext = React.createContext();
@@ -128,6 +130,31 @@ type createStyledOpts = {
   ) => React.ComponentType<any>,
 };
 
+function checkNoopEngine(engine: StandardEngine) {
+  // if no engine provided, we default to no-op, handy for tests
+  // however, print a warning in other envs
+  if (process.env.NODE_ENV !== "test") {
+    engine.noopEngine &&
+      // eslint-disable-next-line no-console
+      console.warn(
+        __DEV__
+          ? `
+Styletron has been switched to a no-op (test) mode.
+
+A Styletron styled component was rendered, but no Styletron engine instance was provided in React context.
+
+Did you forget to provide a Styletron engine instance to React context via using the Styletron provider component?
+
+Note: Providers and Consumers must come from the exact same React.createContext call to work.
+If your app has multiple instances of the "styletron-react" package in your node_module tree,
+your Provider may be coming from a different React.createContext call, which means the styled components
+will not recieve the provided engine instance. This scenario can arise, for example, when using "npm link".
+`
+          : `Styletron Provider is not set up. Defaulting to no-op.`,
+      );
+  }
+}
+
 export function useStyletron() {
   const styletronEngine: StandardEngine = React.useContext(StyletronContext);
   const debugEngine = React.useContext(DebugEngineContext);
@@ -135,6 +162,7 @@ export function useStyletron() {
   const theme = React.useContext(ThemeContext);
   return {
     css: (style: StyleObject) => {
+      checkNoopEngine(styletronEngine);
       const className = driver(style, styletronEngine);
       if (__BROWSER__ && __DEV__ && debugEngine && !hydrating) {
         const debugClassName = debugEngine.debug();
@@ -434,25 +462,7 @@ export function createStyledElementComponent(styletron: Styletron) {
     return (
       <Consumer>
         {(styletron, debugEngine, hydrating) => {
-          {
-            /* if (!styletron) {
-            throw new Error(
-              __DEV__
-                ? `
-A Styletron styled component was rendered, but no Styletron engine instance was provided in React context.
-
-Did you forget to provide a Styletron engine instance to React context via using the Styletron provider component?
-
-Note: Providers and Consumers must come from the exact same React.createContext call to work.
-If your app has multiple instances of the "styletron-react" package in your node_module tree,
-your Provider may be coming from a different React.createContext call, which means the styled components
-will not recieve the provided engine instance. This scenario can arise, for example, when using "npm link".
-`
-                : `Styletron Provider is not correctly setup.`,
-            );
-            return;
-          } */
-          }
+          checkNoopEngine(styletron);
 
           const elementProps = omitPrefixedKeys(props);
           const style = resolveStyle(getInitialStyle, reducers, props);
--- a/packages/styletron-react/src/index.js
+++ b/packages/styletron-react/src/index.js
@@ -147,15 +147,23 @@ export function useStyletron() {
   return [
     function css(style: StyleObject) {
       const className = driver(style, styletronEngine);
-      if (__BROWSER__ && __DEV__ && debugEngine && !hydrating) {
-        const {stack, message} = new Error("stacktrace source");
-        const debugClassName = debugEngine.debug({
-          stackInfo: {stack, message},
-          stackIndex: 1,
-        });
-        return `${debugClassName} ${className}`;
+      if (!(__BROWSER__ && __DEV__)) {
+        return className;
       }
-      return className;
+      const {stack, message} = new Error("stacktrace source");
+      const debugClassName = React.useMemo(
+        () => {
+          if (debugEngine && !hydrating) {
+            return debugEngine.debug({
+              stackInfo: {stack, message},
+              stackIndex: 1,
+            });
+          }
+          return "";
+        },
+        [debugEngine, hydrating],
+      );
+      return debugClassName ? `${debugClassName} ${className}` : className;
     },
   ];
 }
--- a/packages/styletron-react/src/index.js
+++ b/packages/styletron-react/src/index.js
@@ -164,10 +164,16 @@ export function useStyletron() {
     css: (style: StyleObject) => {
       checkNoopEngine(styletronEngine);
       const className = driver(style, styletronEngine);
-      if (__BROWSER__ && __DEV__ && debugEngine && !hydrating) {
-        const debugClassName = debugEngine.debug();
-        return `${debugClassName} ${className}`;
-      }
+      // TODO: Fix/enable debuging mode for this API
+      //
+      // if (__BROWSER__ && __DEV__ && debugEngine && !hydrating) {
+      //   const {stack, stacktrace, message} = new Error("stacktrace source");
+      //   const debugClassName = debugEngine.debug({
+      //     stackInfo: {stack, stacktrace, message},
+      //     stackIndex: 2,
+      //   });
+      //   return `${debugClassName} ${className}`;
+      // }
       return className;
     },
     theme,
--- a/packages/styletron-react/src/index.js
+++ b/packages/styletron-react/src/index.js
@@ -30,12 +30,13 @@ import {addDebugMetadata, DebugEngine} from "./dev-tool.js";
 
 export {DebugEngine};
 
-const StyletronContext = React.createContext<StandardEngine>({
+const noopEngine = {
   renderStyle: () => "",
   renderKeyframes: () => "",
   renderFontFace: () => "",
-  noopEngine: true,
-});
+};
+
+const StyletronContext = React.createContext<StandardEngine>(noopEngine);
 const HydrationContext = React.createContext(false);
 const DebugEngineContext = React.createContext();
 const ThemeContext = React.createContext();
@@ -118,7 +119,7 @@ function checkNoopEngine(engine: StandardEngine) {
   // if no engine provided, we default to no-op, handy for tests
   // however, print a warning in other envs
   if (process.env.NODE_ENV !== "test") {
-    engine.noopEngine &&
+    engine === noopEngine &&
       // eslint-disable-next-line no-console
       console.warn(
         __DEV__
--- a/packages/styletron-react/src/index.js
+++ b/packages/styletron-react/src/index.js
@@ -29,16 +29,25 @@ import {addDebugMetadata, DebugEngine} from "./dev-tool.js";
 
 export {DebugEngine};
 
-const StyletronContext = React.createContext<*>();
+const StyletronContext = React.createContext<StandardEngine>({
+  renderStyle: () => "",
+  renderKeyframes: () => "",
+  renderFontFace: () => "",
+});
 const HydrationContext = React.createContext(false);
 const DebugEngineContext = React.createContext();
+const ThemeContext = React.createContext();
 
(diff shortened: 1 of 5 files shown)


rtsao:
# Theming

Unfortunately, I don't think type inference for themes will be possible in APIs where a `React.createContext` value isn't imported -- the very thing we are trying to avoid in the first place.

```js
const ThemeContext = React.createContext(someTheme);

// $theme type can be inferred
const StyledComponent = styled("div", ({ $theme }) => ({
  color: $theme.textColor
}), ThemeContext);

// theme type can be inferred
function HookComponent() {
  const { theme } useContext(ThemeContext);
  const { css } = useStyletron();
}
```

For consistency with `useStyletron`, potentially the signature of `styled` could be changed to look like:

```js
const StyledComponent = styled("div", (props, [theme, other]) => ({
  color: theme.textColor
}), [ThemeContext, OtherContext]);
```

Of course, the tradeoff is the extra burden of importing these context values around. This is certainly an idiomatic React pattern, but is potentially not worth the burden. That said, I think there is a real benefit (aside from better typing) to moving theming out of core in the sense that there can be multiple themes.

Imagine a web application using two component libraries implemented with Styletron. If theming is in core and both libraries use themes, it is essentially impossible for both of these libraries to coexist without breaking each other or having to know about each other. Of course, multiple themes could be accommodated if different themes were always namespaced in the core theme context, but this is merely convention and has the same drawbacks as legacy React context (naming collisions, etc.)

So realistically, as a convention, theme will probably always need to be namespaced. For example:

```js
const Box = styled("div", ({ $theme }) => ({
  color: $theme.myTheme.textColor
}));
```

So I think this not exactly ideal either. Also, component portability and isolation has always been a paramount concern for Styletron, which is definitely aligned with the approach of theming via vanilla React context.

In summary this is how I see this tradeoff:

## Theming in core

#### Benefits
- Don't have to import React context values when consuming context. Since these are often relative paths, this can be burdensome

#### Drawbacks
- Theme is untyped by default. Type annotations for theme must be manually added, and potentially involves importing a type from elsewhere
- Potential theme collisions. Relies on namespacing conventions to work properly.
- No way to consume arbitrary non-theme context in styles when using `styled`

## React Context-based theming

#### Benefits
- Themes (and other context usage) are automatically inferred with no need for annotations
- No possibility of theme collisions. Components can be completely unaware of different themes.
- Theme is no different than any other context. Straightforward to consume arbitrary non-theme context in styles when using `styled`
- Nothing new to learn with `useStyletron` hook. Themes are just regular context.

#### Drawbacks
- Must import React context values when consuming context. Since these are often relative paths, this can be burdensome

chasestarr:
Do we consider the noop-engine to be a breaking change? It would cause churn right, because many devs rely on the shape in snapshot tests? Not sure of the semver stance here. Those changes generally lgtm, but would be interested to see an isolated pr for it if possible.

I'm on-board with the peer-dep bump. 👍

Can `useStyletron` land without the theme integration? An approach like below should still be possible without theme. Would adding theme support require a major release? I think everyone can agree on the other features included in the pr. If it's easy to add later, it may be worth pulling out for now.

```js
import React from 'react';
import { useStyletron } from "styletron-react";
import MyThemeContext from '../my-theme.js';

export default () => {
  const { css } = useStyletron();
  const { theme } = useContext(MyThemeContext);
  return (
      <div className={css({ color: theme.textColor })}>Blue text</div>
  );
};
```

rtsao:
The React peer dependency change (required for hooks support) is definitely a breaking change. I think adding a default Noop engine is probably technically not breaking, but I think releasing this with some other breaking change makes sense.

I think adding theme support later (depending on the API) can most likely be a minor change.

tajo:
## Unstructured random thoughts

**Great points and the summary of benefits / drawbacks! There is nothing I could really argue with.**

Frankly, this PR is about some minor quality of life improvements. It doesn't fixes bugs or changes things fundamentally. So I think we can't justify breaking some current APIs / patterns just to land this. It would cancel all the goodwill.

> Must import React context values when consuming context. Since these are often relative paths, this can be burdensome

I still see this as a big hurdle. It's already hard to make developers using the `$theme` values exclusively since it's easier to just hardcode `8px` and the `$theme` variable is right there (I am guilty of this too). The extra relative import would add a lot of friction. 

My initial goal was to deprecate `baseui/styled` with this PR since it could be just replaced by Styletron's `styled`. This is already well established pattern

```jsx
const Box = styled("div", ({ $theme }) => ({
  color: $theme.textColor
}));
```

even outside of `baseui` components, so I think at this point it's not realistic to yank the theme out of it and add `useContext` everywhere. That would make devs upset.

Also, the same pattern (`ThemeProvider` / `props.theme`) is used by all other major CSS in JS libs as well, so many developers are already familiar with it. 

## Types

Yea, it seems there is no way how you could type the theme out of the box and indeed, the other libs just go with `any`. 

However, if we remove the assumption that `baseui/styled` should cease to exist, I think we can preserve it as a light wrapper that just adds the types.  So

- by default, the `theme` is typed as `any` and that might be fine if you don't use Flow/TS or just don't care
- if you want to get it typed, you need to re-export `styled` and `useStyletron` so you can use your theme to set the correct type

This also addresses some drawbacks:

> Imagine a web application using two component libraries implemented with Styletron. 

you can namespace your theme

```jsx
<Provider theme={{ baseui: theme }}>
```

and remove the `baseui` part when using `baseui/styled`. Although, this sounds more like a theoretical problem.

## TLDR;
- I really don't want to change the current APIs / patterns. Especially if it doesn't simplify the usage. `theme = useContext()` just adds more friction. Although, I can't argue that it's a cleaner approach.
- If the `styled` function has `props.$theme`, `useStyletron` should provide `theme` as well. They should behave similarly.
- The theme can be typed as `any`, you can reexport `styled` and `useStyletron` to type it properly.
- **I am down to land this PR without the theme support** and keeping this functionality only in `baseui/styled` and `baseui/useStyletron`. Although, I still think it would be valuable to all Styletron users.

rtsao:
### Interface changes

> I really don't want to change the current APIs / patterns.

I would definitely have concerns about usage of `$theme` without namespacing. Do you agree that namespacing the baseUI theme would be a practical necessity as part of a potential migration from `baseui/styled` to vanilla `styled` (in order to avoid collisions)? For example:

```diff
- import {styled} from 'baseui';
+ import {styled} from 'styletron-react';

  const BlueDiv = styled('div', ({$theme}) => ({
-  color: $theme.colors.primary,
+  color: $theme.baseui.colors.primary,
  }));
```

If so, then at the very least, there will have to be a codemod that does the above transformation. 

However, potentially the codemod could instead do:

```diff
- import {styled} from 'baseui';
+ import {styled, ThemeContext} from 'baseui';

- const BlueDiv = styled('div', ({$theme}) => ({
+ const BlueDiv = styled('div', (props, $theme) => ({
    color: $theme.colors.primary,
- }));
+ }), ThemeContext);
```

Standalone, this looks like:
```js
import {styled, ThemeContext} from 'baseui';

const BlueDiv = styled('div', (props, $theme) => ({
  color: $theme.colors.primary,
}), ThemeContext);
```

There's *slightly* more code, but I personally don't think it constitutes a meaningful penalty in terms of boilerplate, especially given the other advantages of an API like this. And assuming there's a codemod, there's very little pain for developers.

Furthermore, there'd be no need for an immediate breakage of the existing interface. It could be deprecated and then removed in the next major version. For example:

```js
import {styled} from "baseui";

// Works, but triggers deprecation warning. Tells developers how to run codemod
const Box = styled("div", ({ $theme }) => ({
  color: $theme.textColor
}));
```

### `useContext` friction

> theme = useContext() just adds more friction

I think this friction could potentially be slightly reduced if `useStyletron` also (optionally) serves as a vanilla context consumer. For example:

```js
import {useStyletron, ThemeContext} from "baseui";

function Component(props) {
  const [css, theme] = useStyletron(ThemeContext);
  /* ... */
}
```

This particular `useStyletron` hook interface would be a potential reason to hold off on releasing it until we sort out theming, since this would be a breaking change from `const {css} = useStyletron();`.

tajo:
All your examples import `from 'baseui'`, is that a typo? Like here:

```jsx
import {styled, ThemeContext} from 'baseui';

const BlueDiv = styled('div', (props, $theme) => ({
  color: $theme.colors.primary,
}), ThemeContext);
```

Otherwise, I am not sure why to expose the `ThemeContext` "binding" if the user is already using `baseui/styled`. We can just do (keep doing) this for them.

>  Do you agree that namespacing the baseUI theme would be a practical necessity as part of a potential migration from baseui/styled to vanilla styled (in order to avoid collisions)?

Yea. That's perfectly fine to do.

**So basically are there two ways forward:**

1. `baseui/styled` stays in place with the same API, `baseui/useStyletron` is introduced and "magically" provides `theme` as well
2. `baseui/styled` is deprecated, everyone should import `styled` and `useStyletron` from `styletron-react` only and pass the `ThemeContext` manually

If we go with `1.` we would still implement this for `styletron-react`:

```js
const [css, theme] = useStyletron(ThemeContext);
const BlueDiv = styled('div', (props, $theme) => ({
  color: $theme.colors.primary,
}), ThemeContext);
```

I am now more inclining to keep offering `baseui` versions of `styled` / `useStyletron` as a simpler alternatives / shortcuts. 

One use-case to thing about: **What if I want to dynamically switch between two themes** (dark mode vs light mode)? Ideally I should be able to make that switch in a single place - the Provider? How would this accommodate that?

```js
import {styled} from 'styletron-react';
import {ThemeContext} from 'baseui';

const BlueDiv = styled('div', (props, $theme) => ({
  color: $theme.colors.primary,
}), ThemeContext);
```

This is also probably the only good reason why not to just import the theme object directly:

```js
import {styled, theme} from 'baseui';

const BlueDiv = styled('div', (props => ({
  color: theme.colors.primary,
}));
```

tajo:
Discussing this with @rtsao 1-1 and here's the conclusion (correct me if wrong!):

We'll do the the most conservative thing and keep the theming out of Styletron for now. This PR will just add this API:

```jsx
const [css] = useStyletron();
```

The array is used so we can potentially return more parameters without a breaking change.

It turns out it makes a lot of sense keeping the theming component library specific so **you can get precise types** and avoid any namespace collisions. Also, it's quite simple these days to use React Context APIs directly (especially with the `useContext` hook), so there's not much benefit to abstracting it by Styletron. Especially not when you consider drawbacks (losing types, namespacing...).

For those who follow [BaseUI](https://github.com/uber-web/baseui), we will keep providing the `styled` function as is and `useStyletron` will be introduced so you can do:

```jsx
import { useStyletron } from 'baseui';

const [css, theme] = useStyletron();
```

packages/styletron-react/src/index.js
```diff
@@ -107,6 +114,52 @@ type createStyledOpts = {
...
+  }
+}
+
+export function useStyletron() {
+  const styletronEngine: StandardEngine = React.useContext(StyletronContext);
+  const debugEngine = React.useContext(DebugEngineContext);
+  const hydrating = React.useContext(HydrationContext);
+  return [
+    function css(style: StyleObject) {
+      checkNoopEngine(styletronEngine);
```
rtsao: Do you think it is necessary for this check every `css` call?  Or should this be hoisted up a level to when `useStyletron` is called.

packages/styletron-react/src/index.js
```diff
@@ -107,6 +114,52 @@ type createStyledOpts = {
...
+      const className = driver(style, styletronEngine);
+      if (__BROWSER__ && __DEV__ && debugEngine && !hydrating) {
+        const {stack, message} = new Error("stacktrace source");
+        const debugClassName = debugEngine.debug({
```
rtsao: I'm wondering if this could be put into `useMemo` so this is only called once (rather than for each `css` call. This might save some unnecessary work.

packages/styletron-react/src/index.js
tajo: good point, moved

packages/styletron-react/src/index.js
tajo: I played with it a bit but no luck. The problem is that it get memoized when `hydrating` or if I remove that from the guarding condition `if (__BROWSER__ && __DEV__ && debugEngine && !hydrating) {}`, then I get the mismatch server/client error. HMR just works (with or w/o the memo). It even fixes the last sourceMap link problem and starts pointing to the right place :)

Screen Shot 2019-04-05 at 3 46 34 PM

Anyway, here's what I ended up with before I gave up:

```jsx
export function useStyletron() {
  const styletronEngine: StandardEngine = React.useContext(StyletronContext);
  const debugEngine = React.useContext(DebugEngineContext);
  const hydrating = React.useContext(HydrationContext);
  checkNoopEngine(styletronEngine);
  return [
    function css(style: StyleObject) {
      const className = driver(style, styletronEngine);
      const debugClassName =
        React.useMemo(
          () => {
            if (__BROWSER__ && __DEV__ && debugEngine) {
              const {stack, message} = new Error("stacktrace source");
              return debugEngine.debug({
                stackInfo: {stack, message},
                stackIndex: 5,
              });
            }
            return "";
          },
          [className],
        ) || "";
      if (debugClassName !== "") {
        return `${debugClassName} ${className}`;
      }
      return className;
    },
  ];
}
```

**It has two problems:**
- It also doesn't add the `debug` className on the initial render (not sure why)
- server/client mismatch warning

Also, not sure if `className` makes completely sense as useMemo's dep in this case. Note that the guarding condition needs to be inside of the memo hook, otherwise React explodes.

rtsao: What about something like:

```js
export function useStyletron() {
  const styletronEngine: StandardEngine = React.useContext(StyletronContext);
  const debugEngine = React.useContext(DebugEngineContext);
  const hydrating = React.useContext(HydrationContext);
  checkNoopEngine(styletronEngine);
  return [
    function css(style: StyleObject) {
      const className = driver(style, styletronEngine);
      // Use ref for error to avoid having stack trace inside of hooks
      const errorRef = useRef(new Error("stacktrace source"));
      const debugClassName = useMemo(
        () => {
          if (__BROWSER__ && __DEV__ && debugEngine && !hydrating) {
            const { stack, message } = errorRef.current;
            return debugEngine.debug({
              stackInfo: { stack, message },
              stackIndex: 1
            });
          }
          return "";
        },
        [hydrating]
      );
      return debugClassName ? `${debugClassName} ${className}` : className;
    }
  ];
}
```

tajo: That works great! Thanks! The PR is updated.

packages/styletron-react/src/index.js
```diff
@@ -30,16 +30,23 @@ import {addDebugMetadata, DebugEngine} from "./dev-tool.js";
...
 export {DebugEngine};
 
-const StyletronContext: Context<any> = createReactContext();
-const HydrationContext = createReactContext(false);
-const DebugEngineContext = createReactContext();
+const StyletronContext = React.createContext<StandardEngine>({
+  renderStyle: () => "",
+  renderKeyframes: () => "",
+  renderFontFace: () => "",
+  noopEngine: true,
```
rtsao: I think rather than duck typing via an extra `noopEngine` property, it may be more clear to assign this object to a variable and then check via reference equality in `checkNoopEngine`.

At first I thought `noopEngine` might have something to do with enabling/disabling this feature (but that'd be a bit nonsensical). Also since this is basically a "private" property, it'd be nice to not have it show up in the `StandardEngine` type.

tajo: Good point! Updated.

PR #305, opened 2019-04-03
Add foundation for devtools

jh3y:
Supersedes https://github.com/styletron/styletron/pull/111

- foundation for devtools
- currently shows applicable atomic classes, defined styles and any extensions for a styled component
devtools-foundation

### To run locally
- `yarn` in root
- `yarn bootstrap`
- `yarn build`
- `cd packages/styletron-react`
- `yarn link`
- `cd ../packages/styletron-engine-atomic`
- `yarn link`
- Run `yarn link styletron-react` and `yarn link styletron-engine-atomic` in your local app repo
- enable `Developer Mode` in Chrome extensions tab
- Load unpacked extension pointing at `packages/devtools-extension`
- Inspect an element in your test app 🎉 

__NOTE:__ Any changes to packages whilst working on the devtools require either an extension refresh in the extensions tab or `yarn build` in the root of the repo to update the deps for a test app 👍


jh3y:
Updated PR to include POC for showing any extensions with `withStyle` and `withStyleDeep`.

extends-integration-styletron

I need to look at `flow` typing and try to fix the issue with that 👍

jh3y:
__UPDATE:__ Here's a gif of using the `devtools-extension` on the included `test-app` 👍 

screencast-gif

tajo:
Note: When testing, I think you need to also do `yarn build` after `yarn bootstrap`

packages/styletron-engine-atomic/src/client/client.js
```diff
@@ -82,9 +83,32 @@ class StyletronClient implements StandardEngine {
     const onNewStyle = (cache, id, value) => {
       const {pseudo, block} = value;
       const sheet: CSSStyleSheet = (this.styleElements[cache.key].sheet: any);
-      const rule = styleBlockToRule(atomicSelector(id, pseudo), block);
+      const selector = atomicSelector(id, pseudo);
+      const rule = styleBlockToRule(selector, block);
       try {
         sheet.insertRule(rule, sheet.cssRules.length);
+        if (__BROWSER__ && window.__STYLETRON_DEVTOOLS__) {
```
tajo: it would be nice to extract this and write some tests

packages/styletron-react/src/index.js
```diff
@@ -77,6 +78,43 @@ class DevProvider extends React.Component<
...
+type StyletronStyles = {
+  classes?: any,
+  styles?: any,
+  extends?: any,
+};
+if (__BROWSER__ && !window.__STYLETRON_DEVTOOLS__) {
+  const atomicMap = new Object();
+  const extensionsMap = new WeakMap();
+  const stylesMap = new WeakMap();
+  const getStyles: StyletronStyles = element => {
```
tajo: also a good candidate for some test?

packages/test-app/src/App.js
```diff
@@ -0,0 +1,83 @@
+/* eslint-env browser */
```
tajo: We probably don't want to add (and maintain) `test-app` in the repo? I usually just `yarn link` things when testing with real apps.

tajo:
It works great! @rtsao can probably can provide better feedback. Can you please resolve the `yarn.lock` conflicts?

It doesn't work for `useStyletron` use case but I guess there's nothing we can do about that.

packages/styletron-react/src/index.js
```diff
@@ -420,6 +480,32 @@ will not recieve the provided engine instance. This scenario can arise, for exam
             elementProps.className = joined;
           }
 
+          if (__BROWSER__ && window.__STYLETRON_DEVTOOLS__) {
```
tajo: Should be this also guarded by `&& __DEV__`? We probably don't want to ship this whole functionality with production bundles, right? Same for other places.

packages/styletron-engine-atomic/src/client/client.js
jh3y: Will look into it 👍

packages/test-app/src/App.js
jh3y: Agreed 👍  I will remove this. I had left this in as a uniform test case for people to try whilst WIP 👍 

I think it could be worthwhile adding a `CONTRIBUTING.md` file that gives contributors a good idea of workflow etc.

packages/styletron-react/src/index.js
jh3y: Yep 👍 Good spot!

jh3y:
Thanks for the feedback @tajo 👍 
I'll look into tests and updating the guards. I've removed the test app and resolved the conflicts.

gergelyke:
@tajo @jh3y do we need anything else here before we can release?

packages/styletron-engine-atomic/src/client/client.js
```diff
@@ -82,9 +83,32 @@ class StyletronClient implements StandardEngine {
     const onNewStyle = (cache, id, value) => {
       const {pseudo, block} = value;
       const sheet: CSSStyleSheet = (this.styleElements[cache.key].sheet: any);
-      const rule = styleBlockToRule(atomicSelector(id, pseudo), block);
+      const selector = atomicSelector(id, pseudo);
+      const rule = styleBlockToRule(selector, block);
       try {
         sheet.insertRule(rule, sheet.cssRules.length);
+        if (__BROWSER__ && __DEV__ && window.__STYLETRON_DEVTOOLS__) {
```
rtsao: I think this might be worth moving into a separate function/file since it's not part of the core functionality.

Sort of how `addDebugMetadata` is abstracted away in `styletron-react`.

packages/styletron-react/src/index.js
```diff
@@ -85,6 +86,43 @@ class DevProvider extends React.Component<
...
+  const extensionsMap = new WeakMap();
+  const stylesMap = new WeakMap();
+  const getStyles: StyletronStyles = element => {
+    const styles: StyletronStyles = {};
+    if (stylesMap.has(element)) {
+      styles.styles = stylesMap.get(element);
+      if (element.classList.length) {
+        const classes = {};
+        for (const className of element.classList) {
+          // $FlowFixMe
```
rtsao: What's the reason for this `$FlowFixMe`? In any case, I think coercion is preferable if it's something not going to be changed ever.

packages/styletron-react/src/index.js
```diff
@@ -472,6 +532,32 @@ export function createStyledElementComponent(styletron: Styletron) {
             elementProps.className = joined;
           }
 
+          if (__BROWSER__ && __DEV__ && window.__STYLETRON_DEVTOOLS__) {
```
rtsao: Same here, I think moving this into a separate function will make this easier to understand.

packages/styletron-react/src/index.js
```diff
@@ -85,6 +86,43 @@ class DevProvider extends React.Component<
...
   __BROWSER__ && __DEV__ ? DevProvider : StyletronContext.Provider;
 
+// DEVTOOLS SETUP
+type StyletronStyles = {
+  classes?: any,
+  styles?: any,
+  extends?: any,
+};
+if (__BROWSER__ && __DEV__ && !window.__STYLETRON_DEVTOOLS__) {
+  const atomicMap = new Object();
```
rtsao: I think this is also worth moving into a separate function `setupDevtools()` and moving into the seame

rtsao: I'm wondering what's the reason to use the `Object` constructor here? 

```suggestion
  const atomicMap = {};
```

Or maybe this should just be a `Map`?

packages/devtools-extension/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "styletron-devtools",
+  "private": true,
+  "version": "0.0.1",
+  "description": "Chrome devtools extension for inspecting Styletron styles",
+  "author": "Jhey Tompkins <jhey@jhey.dev>",
+  "homepage": "https://github.com/styletron/styletron",
+  "repository": "git@github.com:styletron/styletron.git",
+  "bugs": "https://github.com/styletron/styletron/issues",
```
rtsao:
```suggestion
```

packages/devtools-extension/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "styletron-devtools",
+  "private": true,
+  "version": "0.0.1",
+  "description": "Chrome devtools extension for inspecting Styletron styles",
+  "author": "Jhey Tompkins <jhey@jhey.dev>",
+  "homepage": "https://github.com/styletron/styletron",
+  "repository": "git@github.com:styletron/styletron.git",
```
rtsao:
```suggestion
  "repository": "styletron/styletron",
```

packages/devtools-extension/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "styletron-devtools",
+  "private": true,
+  "version": "0.0.1",
+  "description": "Chrome devtools extension for inspecting Styletron styles",
+  "author": "Jhey Tompkins <jhey@jhey.dev>",
+  "homepage": "https://github.com/styletron/styletron",
```
rtsao:
```suggestion
```

jh3y:
Hey @rtsao 👋 

Thanks for the feedback.

I did think about extracting that logic. I wasn't sure where though. I've done it now though.
Logic extracted and `package.json` updated 👍 

Thanks again for the feedback 😃

tajo:
Screen Shot 2019-05-13 at 11 19 32 AM

It seems that classes are always empty now?

jh3y:
All fixed 🛠 👍 

Icons updated with `16`, `48` and `128` sizing 👍

packages/styletron-react/src/index.js
```diff
@@ -1,5 +1,5 @@
 // @flow
-
+/* eslint-env browser */
```
rtsao: Just making note of this for myself.

So it seems `__BROWSER__ && window` is disallowed by the lint rule currently? I'll need to update it so that is allowed.

packages/styletron-react/src/index.js
```diff
@@ -246,7 +256,10 @@ export function withStyleDeep(component, styleArg) {
     addDebugMetadata(styletron, 2);
   }
 
-  return createStyledElementComponent(autoComposeDeep(styletron, styleArg));
+  return createStyledElementComponent(
+    autoComposeDeep(styletron, styleArg),
+    composeExtension(component, styleArg),
```
rtsao: I have a couple thoughts about this:

1. Should this second argument only be provided in development mode? It seems this extension stuff is only used in development, so I assume this is dead code in production?
2. One reason `createStyledElementComponent` has just a single argument is so that it is composable. In comment below, I suggest adding the extra properties needed for the extension to work on the Styletron object itself. If that is feasible, this could look something like:

```js
return createStyledElementComponent(
  addExtension(autoComposeDeep(styletron, styleArg), component)
);
```

I think this keeps the `createStyledElementComponent` implementation dumber/simpler, since it would only be concerned with single styletron objects, the composition of which is handled by other functions (and allowing for arbitrary composition).

packages/styletron-react/src/types.js
```diff
@@ -43,6 +43,16 @@ export type Styletron = {
   },
 };
 
+export type StyletronExtension = {
```
rtsao: This object seems pretty much identical to the `Styletron` object. Given there's just a few additional properties, I'm thinking it might make sense to add them as an optional field to the styletron object (similar to `debug` and `stackIndex`/`stackInfo`). Then `createStyledElementComponent` could still only take one argument and therefore remain composable.

jh3y: Totally 👍

packages/styletron-react/src/index.js
jh3y: Yeah makes perfect sense. I was hoping you might suggest a better way of doing this. I wasn't keen on the way it was being done 👍 I'll make those changes now ⚒

jh3y:
Made some updates @rtsao 👍 

This means `createStyledElementComponent` only accepts one parameter with `Styletron` accepting an optional `ext?:` that contains the extension values.

Tested functionality locally 👍

packages/styletron-react/src/types.js
rtsao: This type is unused now, right? It seems it isn't actually used anywhere.

packages/styletron-react/src/types.js
```diff
@@ -35,14 +35,28 @@ export type Styletron = {
   reducers: Array<ReducerContainer>,
   base: any,
   driver: any,
+  name?: string,
   wrapper: any,
   getInitialStyle: any,
+  ext?: {
+    name?: string,
```
rtsao: Is this `name` property storing the same thing as the `name` property one level up? Or is it some different value. I suppose if it's the same, no need to have it twice? I suppose if `name` is only used by the extension, then it should go under `ext`?

Same thing with `getInitialStyle` and `base`.

jh3y: They aren't storing the same thing as the properties one level up 👍 
They store the values of the component that is being extended. 
It's used at both levels.

jh3y:
Redundant type removed 👍

jh3y:
Dev tools are live 👍 - https://chrome.google.com/webstore/detail/styletron-devtools/bmjdkiphkpkjkicjjgelklhkpbpfiogb

gutenye:
is it available for Firefox?

PR #307, opened 2019-04-15
Add fontFamily fallback support when using declarative @font-face fonts (#236)

### LINKED ISSUE #236: @font-face fallback fonts
Hello!

Using the new `@font-face` feature, great work :) However, ran into an issue today where I wanted to add fallback fonts, but couldn't get it to work.

```
const StyletronComponent = styled('div', {
  fontFamily: {
    src: 'url("/static/fonts/font.woff")',
    fontWeight: 'normal',
    fontStyle: 'normal'
  }
})
```

Let's say I wanted to default back to system fonts, or Arial. Does the current API support this and if so, how?

If not, the resulting `fontFamily: generatedClassName` could be `fontFamily: generatedClassName + ' ' + fallbackFonts`

jh3y:
- Attempts to resolve #236 by catching if a `fontFamily` declaration is an `Array` and processing each entry.
- For example, `fontFamily: [myFontFace, 'cursive']` => `.ae { font-family: ae, cursive }`
- Intercepts within `renderFontFace` to produce a value string from the `Array` 👍

### DIFF
--- a/packages/styletron-engine-atomic/src/client/client.js
+++ b/packages/styletron-engine-atomic/src/client/client.js
@@ -198,25 +198,9 @@ class StyletronClient implements StandardEngine {
     return injectStylePrefixed(this.styleCache, style, "", "");
   }
 
-  renderFontFace(
-    fontFace: FontFaceObject | Array<FontFaceObject | string>,
-  ): string {
-    let result = "";
-    if (Array.isArray(fontFace)) {
-      for (const font of fontFace) {
-        if (typeof font === "object") {
-          const key = declarationsToBlock(fontFace);
-          result += `${this.fontFaceCache.addValue(key, font)},`;
-        } else if (typeof font === "string") {
-          result += `${font},`;
-        }
-      }
-      result = result.slice(0, -1);
-    } else {
-      const key = declarationsToBlock(fontFace);
-      result += this.fontFaceCache.addValue(key, fontFace);
-    }
-    return result;
+  renderFontFace(fontFace: FontFaceObject): string {
+    const key = declarationsToBlock(fontFace);
+    return this.fontFaceCache.addValue(key, fontFace);
   }
 
   renderKeyframes(keyframes: KeyframesObject): string {
--- a/packages/styletron-engine-atomic/src/client/client.js
+++ b/packages/styletron-engine-atomic/src/client/client.js
@@ -198,7 +198,9 @@ class StyletronClient implements StandardEngine {
     return injectStylePrefixed(this.styleCache, style, "", "");
   }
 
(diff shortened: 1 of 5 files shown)


tajo:
This needs some tests too! Something similar to [this](https://github.com/styletron/styletron/blob/master/packages/styletron-engine-atomic/src/client/__tests__/client.browser.js#L120)

packages/styletron-engine-atomic/src/client/client.js
```diff
@@ -199,8 +199,22 @@ class StyletronClient implements StandardEngine {
   }
 
   renderFontFace(fontFace: FontFaceObject): string {
-    const key = declarationsToBlock(fontFace);
-    return this.fontFaceCache.addValue(key, fontFace);
+    let result = "";
+    if (Array.isArray(fontFace)) {
```
chasestarr: based on the `FontFaceObject` type, it is always an [object](https://github.com/styletron/styletron/blob/master/packages/styletron-standard/src/style-types.js#L765). would be best to properly type the expected incoming values

jh3y: I've updated the param type to allow `Array` of `FontFaceObject` and `string`.

packages/styletron-engine-atomic/src/client/client.js
```diff
@@ -198,9 +198,25 @@ class StyletronClient implements StandardEngine {
...
   }
 
-  renderFontFace(fontFace: FontFaceObject): string {
-    const key = declarationsToBlock(fontFace);
-    return this.fontFaceCache.addValue(key, fontFace);
+  renderFontFace(
+    fontFace: FontFaceObject | Array<FontFaceObject | string>,
+  ): string {
+    let result = "";
+    if (Array.isArray(fontFace)) {
```
rtsao: See the above comment, I think this logic instead should go in `renderDeclarativeRules`: https://github.com/styletron/styletron/blob/e1e1ebd8489e0ecea11783dc35f8170a3f8a3f07/packages/styletron-standard/src/index.js#L50-L52

jh3y:
Hey @rtsao 👋 
I moved the logic into `renderDeclarativeRules` and changed the type declaration.
How would the tests look for this? Would `StyletronClient` need to be mocked or imported? There aren't currently any tests in `styletron-standard`. I'd imagine there being a test for `renderDeclarativeRules`?

rtsao:
I suppose an integration test with the atomic engine and styletron-standard might be the best thing here. This doesn’t quite exist yet but would probably be similar to the existing hydration tests

jh3y:
I've added a basic `driver` test within `styletron-standard` package.

Tests `getCss` output against expected `String` after invoking the `driver` function 👍

PR #309, opened 2019-04-15
breaking: Replace $ref by React.forwardRef

tajo:
**This is a breaking change**. Motivation: 

- `$ref` was invented before native solution `React.forwardRef` existed
-   it removes one of the significant pitfalls since simple `ref` prop will just work after this change

Update for consumers should be as simple as doing `r/$ref/ref` globally. Afaik, there is no reason why would anyone wanted to get `ref` of the wrapper component (it doesn't even really work?).

Also, the React team has an intent to drop `React.forwardRef` and forward by default.

EDIT: [Apps](https://github.com/uber-web/baseui/blob/master/src/styles/styled.js#L14) implementing custom `styled` with `createStyled` might need to add additional `React.forwardRef` (the default wrapper is an identity function).

### DIFF
--- a/packages/styletron-react/src/__tests__/tests.browser.js
+++ b/packages/styletron-react/src/__tests__/tests.browser.js
@@ -288,7 +288,7 @@ test("$-prefixed props not passed", t => {
   );
 });
 
-test("$ref", t => {
+test("ref forwarding", t => {
   t.plan(1);
 
   const Widget = styled("button", {color: "red"});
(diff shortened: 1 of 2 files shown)


chasestarr:
tested and works. does it make sense to provide a warning in dev for if someone attempts to use the `$ref` prop? (thinking about how to best communicate this change)

rtsao:
> does it make sense to provide a warning in dev for if someone attempts to use the $ref prop? (thinking about how to best communicate this change)

I think that for this release it would be nice if `$ref` was deprecated (usage triggers a warning message) but still worked. Then in the next major version we'd remove it entirely.

tajo:
@chasestarr @rtsao Updated.

PR #317, opened 2019-05-30
Fix forwarding of React.createRef() refs

tajo:
Refs created with `React.createRef()` currently break `styletron-react` since they are not forwarded as functions into the devtool instrumentation. This PR fixes it and adds tests. Fortunately, legacy string refs are being transformed info functions by `React.forwardRef` so no change needed there. 

```js
if (ref) {
  if (typeof ref === "function") {
    return ref(element);
  }
  ref.current = element;
  return ref;
}
```

It looks a bit iffy but it seems this is the [only way](https://github.com/facebook/react/issues/13029) to compose refs created by `createRef()`.

cc @jh3y

### DIFF
--- a/packages/styletron-react/src/__tests__/tests.browser.js
+++ b/packages/styletron-react/src/__tests__/tests.browser.js
@@ -337,7 +337,7 @@ test("$-prefixed props not passed", t => {
   );
 });
 
-test("ref forwarding", t => {
+test("callback ref forwarding", t => {
   t.plan(1);
 
   const Widget = styled("button", {color: "red"});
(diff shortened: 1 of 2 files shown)


packages/styletron-react/src/__tests__/tests.browser.js
```diff
@@ -368,6 +368,64 @@ test("ref forwarding", t => {
   Enzyme.mount(<TestComponent />);
 });
 
+test("React.createRef() ref forwarding", t => {
+  t.plan(1);
+
+  const Widget = styled("button", {color: "red"});
+  class TestComponent extends React.Component<{}> {
+    widgetInner: {current: null | HTMLButtonElement};
```
rtsao: Flow seems unhappy with this.

tajo: Did a fix but not great. Not entirely sure how to type the forwardRef thing (which seems to be the real problem).

PR #318, opened 2019-05-31
Fix refs again

tajo:
We need to detect if underlying component can deal with ref aka:

- it forwards them
- it is a class component
- `string` like `div` or `button` is being passed / created

Also, `$ref` was never passed through in the DEV mode.

### DIFF
--- a/packages/styletron-react/src/dev-tool.js
+++ b/packages/styletron-react/src/dev-tool.js
@@ -1,7 +1,7 @@
 /* eslint-env browser */
 /* global module */
 import * as React from "react";
-import * as ReactIs from "react-is";
+import {isForwardRef} from "react-is";
 
 export function addDebugMetadata(instance, stackIndex) {
   const {stack, stacktrace, message} = new Error("stacktrace source");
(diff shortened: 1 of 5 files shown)


packages/styletron-react/src/dev-tool.js
```diff
@@ -1,5 +1,7 @@
 /* eslint-env browser */
 /* global module */
+import * as React from "react";
+import * as ReactIs from "react-is";
```
rtsao: This is only in development, but in theory this could still be tree-shaken so I think it's probably worth importing the specific export

```suggestion
import {isForwardRef} from "react-is";
```

packages/styletron-react/src/dev-tool.js
```diff
@@ -9,6 +11,19 @@ export function addDebugMetadata(instance, stackIndex) {
   };
 }
 
+export function canAcceptRef(Component) {
+  if (typeof Component === "string") {
+    return true;
+  }
+  if (ReactIs.isForwardRef(<Component />)) {
```
rtsao:
```suggestion
  if (isForwardRef(<Component />)) {
```

PR #320, opened 2019-06-11
Improve typing of styletron-react internals

rtsao:
The reducer container type is a disjoint union to allow for refinements.

That said, internal type safety might not be worth it for the runtime cost of this check. At the very least, this is a good starting place for doing a more limited, precise casting (as opposed to `any`).

### DIFF
--- a/packages/styletron-react/src/index.js
+++ b/packages/styletron-react/src/index.js
@@ -20,7 +20,6 @@ import type {
   ReducerContainer,
   AssignmentCommutativeReducerContainer,
   NonAssignmentCommutativeReducerContainer,
-  Reducer,
   StyledFn,
   WithStyleFn,
   WithTransformFn,
@@ -377,10 +376,10 @@ export function createDeepMergeReducer(
 
 export function composeStatic(
   styletron: Styletron,
-  reducerContainer: ReducerContainer,
+  reducerContainer: AssignmentCommutativeReducerContainer,
 ) {
   if (styletron.reducers.length === 0) {
-    const style = reducerContainer.reducer(styletron.getInitialStyle(), {});
+    const style = reducerContainer.reducer(styletron.getInitialStyle());
     const result: Styletron = {
       reducers: styletron.reducers,
       base: styletron.base,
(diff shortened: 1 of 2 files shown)


packages/styletron-react/src/index.js
```diff
@@ -20,7 +20,6 @@ import type {
   ReducerContainer,
   AssignmentCommutativeReducerContainer,
   NonAssignmentCommutativeReducerContainer,
```
chasestarr: when looking through yesterday, I noticed the `NonAssignmentCommutativeReducerContainer` type was not used - should it be maintained, or leveraged in this file?

rtsao: Yeah, `NonAssignmentCommutativeReducerContainer` is not used directly so there's no need to import it.

However, the type is inferred in [`composeDynamic`](https://github.com/styletron/styletron/pull/320/commits/5847117a3551043fd3f79a531f5ce429be34adf3#diff-f723eef49306ed376aa6deab2dcb9205L432)(Flow is able to refine the `ReducerContainer` disjoint union via the `assignmentCommutative` property).

packages/styletron-react/src/index.js
```diff
@@ -559,7 +557,10 @@ export function resolveStyle(
   let result = getInitialStyle();
   let i = reducers.length;
   while (i--) {
-    result = reducers[i].reducer(result, props);
+    const reducer = reducers[i];
+    result = reducer.assignmentCommutative
```
rtsao: This will incur some (potentially insignificant) runtime penalty. Might worth some casting to avoid it.

packages/styletron-react/src/index.js
chasestarr: right, does the type need to be in scope for it to be inferred?

packages/styletron-react/src/index.js
chasestarr: we could add a default value to `props` of an empty object. this way, `reducer.reducer` can always be called with both arguments; removing the need for a runtime check

chasestarr: I prefer to typecast since we're only satisfying flow in this case, is isolated to the utility function, and we know that calling with additional arguments is a safe action. The main issue is if the two reducer types significantly diverge. I tested locally with the def below passing, and even changed the reducer type if the container types to see if an error would be surfaced (which it did). What's your preference? 

```js
const reducer = (reducers[i].reducer: (StyleObject, Object) => StyleObject);
result = reducer(result, props);
```

rtsao: Thanks, I like your suggestion. Just commited that

PR #321, opened 2019-06-12
Use className instead of element as the identifier for Chrome extension

tajo:
In the dev mode, `styletron-react` creates a global data-structure (map) that is consumed by our Chrome extension. It maps elements (DOM nodes) to the styles meta-data. In order to get DOM nodes in styletron-react, we use [refs](https://reactjs.org/docs/refs-and-the-dom.html). Unfortunately, it causes multiple issues.

It seems we can just use `className` attribute as our key instead of the element since there's a direct relation between that and resulting styles. That means no refs and messy workarounds!

There is one edge-case that can happen: Two different styled components having different initial styles but resulting into the same style-set because of `withStyle` overrides. Then, you could get a partly incorrect info in the devtools. However, if you also use the `debug` mode (enabled by default in most of apps), the className also includes the `__debugX` class that is always unique and prevents this edge-case. So I think we don't have to worry about it. 

I've also added test for `React.useRef()` forwarding.

### DIFF
--- a/packages/devtools-extension/devtools_script.js
+++ b/packages/devtools-extension/devtools_script.js
@@ -4,15 +4,18 @@ const elementsPanel = panels.elements;
 
 elementsPanel.createSidebarPane("Styletron", sidebar => {
   elementsPanel.onSelectionChanged.addListener(() => {
-    inspectedWindow.eval("__STYLETRON_DEVTOOLS__.getStyles($0)", (res, err) => {
-      if (err && err.isError) {
-        throw new Error(`Styletron devtools: ${err.description}`);
-      }
-      if (res) {
-        sidebar.setObject(res, "Styletron Styles");
-      } else {
-        sidebar.setObject(null, "Not a styled element");
-      }
-    });
+    inspectedWindow.eval(
+      "__STYLETRON_DEVTOOLS__.getStyles($0.className)",
+      (res, err) => {
+        if (err && err.isError) {
+          throw new Error(`Styletron devtools: ${err.description}`);
+        }
+        if (res) {
+          sidebar.setObject(res, "Styletron Styles");
+        } else {
+          sidebar.setObject(null, "Not a styledddd element");
+        }
+      },
+    );
   });
 });
--- a/packages/devtools-extension/devtools_script.js
+++ b/packages/devtools-extension/devtools_script.js
@@ -13,7 +13,7 @@ elementsPanel.createSidebarPane("Styletron", sidebar => {
         if (res) {
           sidebar.setObject(res, "Styletron Styles");
         } else {
-          sidebar.setObject(null, "Not a styledddd element");
+          sidebar.setObject(null, "Not a styled element");
         }
       },
     );
--- a/packages/styletron-react/src/dev-tool.js
+++ b/packages/styletron-react/src/dev-tool.js
@@ -1,7 +1,5 @@
 /* eslint-env browser */
 /* global module */
-import * as React from "react";
-import {isForwardRef} from "react-is";
 
 export function addDebugMetadata(instance, stackIndex) {
   const {stack, stacktrace, message} = new Error("stacktrace source");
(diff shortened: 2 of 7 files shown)


packages/styletron-react/src/dev-tool.js
```diff
@@ -55,21 +17,25 @@ type StyletronStyles = {
 };
 export const setupDevtoolsExtension = () => {
   const atomicMap = {};
-  const extensionsMap = new WeakMap();
-  const stylesMap = new WeakMap();
-  const getStyles: StyletronStyles = element => {
+  const extensionsMap = new Map();
+  const stylesMap = new Map();
```
rtsao: Any concerns about these maps growing indefinitely? The nice thing about the `WeakMaps` was if elements get garbage collected, so does the metadata.

tajo:
> Offline you mentioned a [`__reactInternalInstance$xyz` property](https://github.com/bvaughn/react-devtools-experimental/blob/6c8ae706ee22e6a3f69940be0a51cc9763a3a89f/src/backend/views/Overlay.js#L288-L290) existing on DOM nodes that points to the corresponding `React.FiberNode` instance.
> 
> How would using that compare to this approach? I guess we would stash the metadata on that fiber node itself using a `WeakMap`?

Well, I am actually not sure how to "stash the metadata on that fiber node".  It seems that Fiber node is a "unit of work" rather than a unique instance of specific component. It might not be possible (without refs)? `react-experimental-devtools` uses that API to only get a `displayName`.

packages/devtools-extension/devtools_script.js
```diff
@@ -4,15 +4,18 @@ const elementsPanel = panels.elements;
...
+    inspectedWindow.eval(
+      "__STYLETRON_DEVTOOLS__.getStyles($0.className)",
+      (res, err) => {
+        if (err && err.isError) {
+          throw new Error(`Styletron devtools: ${err.description}`);
+        }
+        if (res) {
+          sidebar.setObject(res, "Styletron Styles");
+        } else {
+          sidebar.setObject(null, "Not a styledddd element");
```
chasestarr:
```suggestion
          sidebar.setObject(null, "Not a styled element");
```

packages/styletron-react/src/dev-tool.js
tajo: The same thing already happens with the `MultiCache` right? We never delete/remove unused styles. 

Since this is enabled only in the dev mode and resets with refreshing, it should be fine? If you are creating thousands of styled elements / using styletron to animate something, you would run into bigger problems even without this.

rtsao: True, although that cache is indexed on atomic styles, which I would expect to grow at a much slower rate than unique class strings. In the worst (albeit totally unrealistic) case, given `n` unique atomic styles there are `n!` possible class strings.

I don't expect this to be a problem in practice, but just something to think about.

rtsao:
> > Offline you mentioned a [`__reactInternalInstance$xyz` property](https://github.com/bvaughn/react-devtools-experimental/blob/6c8ae706ee22e6a3f69940be0a51cc9763a3a89f/src/backend/views/Overlay.js#L288-L290) existing on DOM nodes that points to the corresponding `React.FiberNode` instance.
> > How would using that compare to this approach? I guess we would stash the metadata on that fiber node itself using a `WeakMap`?
> 
> Well, I am actually not sure how to "stash the metadata on that fiber node". It seems that Fiber node is a "unit of work" rather than a unique instance of specific component. It might not be possible (without refs)? `react-experimental-devtools` uses that API to only get a `displayName`.

I have a simple proof of concept that demonstrates an approach I think could possibly work:
https://codesandbox.io/embed/distracted-shannon-p2jgp

rtsao:
Since the existing version is buggy, let's publish this in its current state (and can maybe look at improved solutions later)

tajo:
The new extension is published/pending for review, should be up in an hour. 

(mismatch of styletron-react and extension doesn't bomb the page, it just doesn't display anything, should not be a big deal)

PR #323, opened 2019-06-17
Replace vendored StyleObject with types imported from csstypes

### LINKED ISSUE #228: Use csstype as a dependency

chasestarr:
closes #228

- Imports separate types that make up the full `Properties` type because only the `StandardProperties` in the intersection was being included. Caught by [this](https://github.com/styletron/styletron/blob/master/packages/flow-type-tests/inferred-static.js#L9) test which included the `gridGap` property only available in the `ObsoleteProperties` type. See approach to fix this [here](https://github.com/chasestarr/styletron/blob/df81702d798ff4a68f3e36866ad27c924f3ef813/packages/styletron-standard/src/style-types.js#L27)
- I believe I included all of the styletron-specific type customizations (fonts and animations), but let me know if something was skipped.
- The `style-types.js` file no longer exports all of the various composite types. Is that something we still need?


packages/styletron-standard/package.json
```diff
@@ -30,6 +30,7 @@
     "prepublish": "npm run build"
   },
   "dependencies": {
+    "csstype": "^2.6.5",
```
rtsao:
```suggestion
    "csstype": "2.6.5",
```

chasestarr:
Getting errors when yarn link-ing this change locally, but not sure if I'm doing it correctly with lerna here. Found a neat flag `--profile` that gives a bit more information on what's happening. Planning to investigate.

chasestarr:
figured out the linking and this branch works well

PR #325, opened 2019-06-20
cleanup READMEs

tajo:
Deleted the most of READMEs and added more links to styletron.org since the information was getting outdated and no need to have it in two places.


README.md
```diff
@@ -2,142 +2,44 @@
...
 
-[*Looking for v3.x docs?*](https://www.styletron.org/v3/) | [*v3.x to v4.x migration guide*](docs/v3-migration-guide.md)
-
-## Packages
-
-* [`styletron-engine-atomic`](packages/styletron-engine-atomic)
-* [`styletron-react`](packages/styletron-react)
-* [`styletron-standard`](packages/styletron-standard)
-
-## Design principles
```
rtsao: These aren't on the styletron.org site, are they?

Something like this I think is potentially valuable to keep in the README

tajo: Yea, it's not exactly there or described differently. I put this section back.

PR #331, opened 2019-07-23
feat(type): allow null StyleObject

leyanlo:
null can be used to clear values when restyling a styled component, and should be considered a valid type


rtsao:
Closing in favor of https://github.com/styletron/styletron/pull/346

PR #335, opened 2019-09-04
Sort client side media queries correctly

tajo:
This bug happens in one of our apps. The media query sorting works only for SSRed rules. If you add a new and unique media query on the client, it just gets appended to the top. There are two different problems:

- `MultiCache.sortedCacheKeys` array is not correctly hydrated (it appears empty to the client)
-  even when `sortedCacheKeys` is fixed, `this.container.insertBefore()` is too naive and doesn't really assume that `container` could have elements other than Styletron's `<style>` tags but that's almost always the case when `document.head` is the `container`

With this PR, the insert logic is more robust - it goes through `container.children` and compares `media` attributes (instead of blindly passing "insert index").

`keyframes` are moved to the top of stylesheets to make the media query sorting logic simpler - we append max media queries to the end of container and not worry about keyframes, technically it doesn't matter but it looks weird to have a keyframe tag in the middle of media query tags. Also, it doesn't really matter since keyframes don't compete with specificity so they can be placed anywhere.

Additionally, this PR also updates all eslint/prettier stuff and doesn't use eslint to run prettier anymore (still checks for the validity though). It seems the latest VS Code integration is not able to pick up the prettier settings through the eslint. There's just better support for this setup.

### DIFF
--- a/packages/styletron-engine-atomic/src/cache.js
+++ b/packages/styletron-engine-atomic/src/cache.js
@@ -6,7 +6,7 @@ import sortMq from "./sort-css-media-queries.js";
 export class MultiCache<T> {
   caches: {[string]: Cache<T>};
   idGenerator: SequentialIDGenerator;
-  onNewCache: (string, Cache<T>, string) => any;
+  onNewCache: (string, Cache<T>, ?string) => any;
   onNewValue: (cache: Cache<T>, id: string, value: T) => any;
   sortedCacheKeys: string[];
 
@@ -32,7 +32,7 @@ export class MultiCache<T> {
       const insertBeforeMedia =
         keyIndex < this.sortedCacheKeys.length - 1
           ? this.sortedCacheKeys[keyIndex + 1]
-          : "";
+          : void 0;
       this.caches[key] = cache;
       this.onNewCache(key, cache, insertBeforeMedia);
     }
--- a/packages/styletron-engine-atomic/src/cache.js
+++ b/packages/styletron-engine-atomic/src/cache.js
@@ -6,7 +6,7 @@ import sortMq from "./sort-css-media-queries.js";
 export class MultiCache<T> {
   caches: {[string]: Cache<T>};
   idGenerator: SequentialIDGenerator;
-  onNewCache: (string, Cache<T>, number) => any;
+  onNewCache: (string, Cache<T>, string) => any;
   onNewValue: (cache: Cache<T>, id: string, value: T) => any;
   sortedCacheKeys: string[];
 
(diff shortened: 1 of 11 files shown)


packages/styletron-engine-atomic/src/client/client.js
```diff
@@ -222,3 +228,18 @@ class StyletronClient implements StandardEngine {
 }
 
 export default StyletronClient;
+
+function findSheetIndexWithMedia(children, media) {
```
rtsao: I refactored this logic in https://github.com/styletron/styletron/pull/335/commits/d1ab18a616cdb446571a78721515aff91168202e to avoid `Array.from` and `Array.prototype.findIndex` which would be reliant upon ES2015 polyfills (not merely Babel transpilation).

packages/styletron-engine-atomic/src/client/client.js
```diff
@@ -222,3 +228,18 @@ class StyletronClient implements StandardEngine {
...
+    const child = children[index];
+    if (
+      child.tagName === "STYLE" &&
+      ((child: any): HTMLStyleElement).media === media
+    ) {
+      return index;
+    }
+  }
+
+  return -1;
```
rtsao: `Array.prototype.findIndex` returns `-1` if not found, so I've preserved this here.

Should we handle the case that it doesn't exist for some reason? This should never happen, but if it did, we'd attempt array access at `-1` which would throw.

packages/styletron-engine-atomic/src/cache.js
```diff
@@ -28,9 +28,13 @@ export class MultiCache<T> {
       cache.key = key;
       this.sortedCacheKeys.push(key);
       this.sortedCacheKeys.sort(sortMq);
-      const insertAtIndex = this.sortedCacheKeys.indexOf(key);
+      const keyIndex = this.sortedCacheKeys.indexOf(key);
+      const insertBeforeMedia =
+        keyIndex < this.sortedCacheKeys.length - 1
+          ? this.sortedCacheKeys[keyIndex + 1]
+          : "";
```
rtsao: Using `""` as the `insertBeforeMedia` value to trigger "append to container" logic seems a bit odd to me given that sheets with no media queries (i.e. `media === ""`) are ordered *before* sheets that do have media queries.

Perhaps it'd be better to use `undefined` instead to indicate the absence of a `insertBeforeMedia` value, which implies the sheet should not go before anything else and hence last.

packages/styletron-engine-atomic/src/client/client.js
tajo: This is still fine even with `-1`. `children[-1]` returns `undefined` and `insertBefore(element, undefined)` just appends the element to the end:

> If the reference node is null, the specified node is added to the end of the list of children of the specified parent node.

rtsao: Ah, TIL

packages/styletron-engine-atomic/src/client/client.js
```diff
@@ -110,17 +110,22 @@ class StyletronClient implements StandardEngine {
     // Setup style cache
     this.styleCache = new MultiCache(
       styleIdGenerator,
-      (media, _cache, insertAtIndex) => {
+      (media, _cache, insertBeforeMedia) => {
         const styleElement = document.createElement("style");
         styleElement.media = media;
-        if (insertAtIndex >= this.container.children.length) {
+        if (typeof insertBeforeMedia === "undefined") {
```
rtsao:
```suggestion
        if (insertBeforeMedia === void 0) {
```

tajo: typeof check should be safe anyway?

Issue #337, opened 2019-09-19
Cannot use DebugEngine with Typescript


koukalp:
I am following the documentation when setting up this awesome library with React and Typescript. 

1. Import the dependencies: 
```
import {Provider as StyletronProvider, DebugEngine} from "styletron-react";
```

2. init the debugger: 
```
const debug = process.env.NODE_ENV === "production" ? void 0 : new DebugEngine();
```

3. Typescript gives a warning on the `new DebugEngine()`
![image](https://user-images.githubusercontent.com/41000441/65223445-836cce00-dac1-11e9-9faa-45ad8ac6ceb0.png)

4. in step 1, import `BrowserDebugEngine` instead:
```
import {BrowserDebugEngine,Provider as StyletronProvider} from "styletron-react";
```

5. use it in step 2: 
```
const debug =process.env.NODE_ENV === "production" ? void 0 : new BrowserDebugEngine();
```

6. React app won't load with the following error: 
![image](https://user-images.githubusercontent.com/41000441/65223676-fbd38f00-dac1-11e9-896d-64c1d54a6a18.png)

The debug feature would be quite useful. Everything else works flawlessly. Would be great if smbd could give me a hint on how to fix my issue

choyongjoon:
I'm using `// @ts-ignore` to ignore this error.

koukalp:
thanks for answer. That doesn't really fix anything for me: 

```
// @ts-ignore
  process.env.NODE_ENV === "production" ? void 0 : new BrowserDebugEngine();
```

chasestarr:
This repo only maintains and publishes flow-types. I suggest sending a PR to [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/styletron-react) if you find it helpful

PR #339, opened 2019-09-23
typescript type definitions

zxbodya:
While building a [tool for flow to typescript migration](https://github.com/zxbodya/flowts), I used styletron as one of the projects to test it. And so - now I have typescript version of styletron(https://github.com/zxbodya/styletron/tree/ts)

Type definitions in this PR are generated from that typescript version.

While doing it, I was trying to make it compatible with definitions currently in DefinitelyTyped(ensuring things like type parameters order in generics to be the same), also I did some additional corrections while using it with migrated version of baseui

For the most part(except maybe some places where flow types were not very accurate) - new type definitions are more accurate comparing to the version in DefinitelyTyped.

btw, this should also help with #337

### DIFF
--- a/packages/styletron-react/src/index.d.ts
+++ b/packages/styletron-react/src/index.d.ts
@@ -1,0 +1,53 @@
+declare global {
+    interface Window {
+        __STYLETRON_DEVTOOLS__: any;
+    }
+}
+import * as React from "react";
+import { driver, StandardEngine, StyleObject } from "styletron-standard";
+import { Styletron, ReducerContainer, AssignmentCommutativeReducerContainer, StyledFn, WithStyleFn, WithTransformFn, WithWrapperFn, StyletronProps } from "./types";
+import { DebugEngine } from "./dev-tool";
+export { DebugEngine };
+export { StyleObject };
+export { StyletronProps };
+declare type DevProviderProps = {
+    children: React.ReactNode;
+    value: StandardEngine;
+    debugAfterHydration?: boolean;
+    debug?: any;
+};
+declare class DevProvider extends React.Component<DevProviderProps, {
+    hydrating: boolean;
+}> {
+    constructor(props: DevProviderProps);
+    componentDidMount(): void;
+    render(): JSX.Element;
+}
+export declare const Provider: typeof DevProvider | React.ProviderExoticComponent<React.ProviderProps<StandardEngine>>;
+export declare function DevConsumer(props: {
+    children: (c: any, b: any, a: any) => React.ReactNode;
+}): JSX.Element;
+declare type createStyledOpts = {
+    getInitialStyle: () => StyleObject;
+    driver: typeof driver;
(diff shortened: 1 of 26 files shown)


packages/styletron-react/src/index.d.ts
```diff
@@ -0,0 +1,51 @@
...
+import * as React from "react";
+import { driver, StandardEngine, StyleObject } from "styletron-standard";
+import { Styletron, ReducerContainer, AssignmentCommutativeReducerContainer, StyledFn, WithStyleFn, WithTransformFn, WithWrapperFn, StyletronProps } from "./types";
+import { DebugEngine } from "./dev-tool";
+export { DebugEngine, StyletronProps };
+declare type DevProviderProps = {
+    children: React.ReactNode;
+    value: StandardEngine;
+    debugAfterHydration?: boolean;
+    debug?: any;
```
choyongjoon:
```suggestion
    debug?: BrowserDebugEngine | NoopDebugEngine;
```

packages/styletron-react/src/index.d.ts
```diff
@@ -0,0 +1,51 @@
...
+declare class DevProvider extends React.Component<DevProviderProps, {
+    hydrating: boolean;
+}> {
+    constructor(props: DevProviderProps);
+    componentDidMount(): void;
+    render(): JSX.Element;
+}
+export declare const Provider: typeof DevProvider | React.ProviderExoticComponent<React.ProviderProps<StandardEngine>>;
+export declare function DevConsumer(props: {
+    children: (c: any, b: any, a: any) => React.ReactNode;
```
choyongjoon:
```suggestion
    children: (
        styletronEngine: StandardEngine,
        debugEngine: BrowserDebugEngine | NoopDebugEngine,
        hydrating: boolean
    ) => React.ReactNode;
```

zxbodya:
> There are many `any`s compared to DefinitelyTyped's one. How about to check all `any`s and get rid of them as many as possible?

that would be great to do this

but, it is kind of outside of the scope of what I was doing - my goal was more about to convert as much of existing flow types as possible, and to reveal pain points in converted code by trying to do fixes needed after migration from flow(btw, I also have converted baseui codebase, to reveal what additional fixes in styletron needed to support usages in baseui). It was not really a goal to provide complete type coverage… and things which are `any` are actually not covered in flow types

however, if you feel like doing it - you can try doing it on top of ts branch (https://github.com/zxbodya/styletron/tree/ts) - I would be happy to update this pr to include those changes, also btw - it might be even better adding it as flow type annotations in sources (it is relatively easy to update ts version to include changes in flow sources)

other option about doing it(which I am planning to try), would be to automate generating `d.ts` files and rebasing fixes on top of generated files - then it would be easier to backport types from definitely typed(and also to require less effort to create typescript typings from existing sources with flow type annotations)

tajo:
new PR based of this one: https://github.com/styletron/styletron/pull/371

Issue #176, opened 2017-10-17
create jest-styletron similar to (and for the same reasons as) jest-styled-components


koulmomo:
When running jest snapshot tests against styletron styled components, we encounter the same issues as what happened when people tried doing the same with [styled-components](https://www.styled-components.com/) (see: https://github.com/styled-components/jest-styled-components/issues/29).

clayne11:
This solution didn't work for me. I think the correct solution is to create a custom serializer for Jest that filters out the `$theme` and some of the other internal props.

tajo:
I think we can elegantly solve this by using the [No-op engine](https://github.com/styletron/styletron/issues/256) for testing. 

If you use the real engine, we can assume that you actually want to snapshot classNames / styles and you are fine with (possibly) frequent updates. Also, the `jest-styled-components` issue mentions problems when using Button across multiple components causing massive churn in snapshot tests. That can be solved by using shallowRendering when taking snapshots.

PR #342, opened 2019-10-11
Add styletron-engine-snapshot package

koulmomo:
The styletron atomic-engine [auto-generates atomic css classes for you](https://www.styletron.org/#performance) making sure to minimize the number of classnames + style declarations when rendering.

This is great for production as it ensures we only ever send the minimal amount of css necessary for render [and stop css size growth as we keep adding features](https://developers.facebook.com/videos/2019/building-the-new-facebookcom-with-react-graphql-and-relay/) (28:00), but it makes snapshot testing components brittle and adds noise to diffs.

Snapshots generated with `styletron-engine-atomic` have 2 main issues

1. non-debuggable classNames in snapshots ("ae af ")
2. className generation is non-deterministic. If you create a new component,
it might break an unrelated snapshot since that component rendering shifts the
generation for the old snapshot.

This was first brought up in #176 which ended in a recommendation to use the [no-op engine](https://github.com/styletron/styletron/issues/256)
in testing so that snapshots do not contain any class/style information.

Using the no-op engine means that snapshots lose information about classNames/styles, and only contain information about behavior and/or prop passing.
This package attempts to provide a better solution for this problem.

The package provides a *deterministic* engine that simply returns a `JSON.stringify`-ed version of the style object (with alpha-sorted keys).
A snapshot should not fail **UNLESS** a style applied to the component changes (which you want).
It also makes it easier to detect what exactly changed in the css in the snapshot since we actually now render all of the css values (we are essentially almost rendering the styles inline).

## Downsides of This Package/Approach

1. The engine generates invalid css class names (as they are essentially JSON)
2. Styles are **NOT ACTUALLY APPLIED**, so if you have existing tests that check/rely on computed styles, they no longer work.

## Future Improvements

There is a path forward that can fix the downsides mentioned above.
We can modify the engine into creating valid class names + style declerations **PER-STYLE**, similar to what https://acss.io/ does.
That would make it so that snapshots still diff nicely, while actually applying styles as well.


packages/styletron-engine-snapshot/src/create-render.js
```diff
@@ -0,0 +1,41 @@
...
+2) className generation is non-deterministic. If you create a new component,
+it might break an unrelated snapshot since that component rendering shifts the
+generation for the old snapshot.
+
+This adds a "deterministic" engine that simply returns a `JSON.stringify`-ed
+version of the style object (with aplha-sorted keys). This makes it so that a
+snapshot should not fail **UNLESS** a style applied to the component changes (which you want).
+It also makes it easier to detect **what** exactly changed in the css in the snapshot since we
+actually now render all of the css values (we are essentially almost rendering the styles inline).
+*/
```
tajo: Would add this description in to the README + also stating that it output non-executable (invalid) code.

koulmomo: updated

packages/styletron-engine-snapshot/src/create-render.js
```diff
@@ -0,0 +1,41 @@
...
+export default function createRender(kind: string) {
+  return function render(
+    obj: ?(StyleObject | FontFaceObject | KeyframesObject),
+  ): string {
+    if (!obj) {
+      return "";
+    }
+
+    return `${kind}=${stringify(obj, {space: 2})
+      .replace(/["]/g, "")
```
rtsao: Should this escape quotes instead?

I'd imagine a change like this should yield a snapshot change:
```diff
{
  ":after": {
-   content: "Hello world" // Broken
+   content: `"Hello world"`
  }
}
```

yarn.lock
```diff
@@ -4187,6 +4187,13 @@ json-stable-stringify-without-jsonify@^1.0.1:
   resolved "https://registry.yarnpkg.com/json-stable-stringify-without-jsonify/-/json-stable-stringify-without-jsonify-1.0.1.tgz#9db7b59496ad3f3cfef30a75142d2d930ad72651"
   integrity sha1-nbe1lJatPzz+8wp1FC0tkwrXJlE=
 
+json-stable-stringify@^1.0.1:
+  version "1.0.1"
+  resolved "https://unpm.uberinternal.com/json-stable-stringify/-/json-stable-stringify-1.0.1.tgz#9a759d39c5f2ff503fd5300646ed445f88c4f9af"
```
rtsao: The lockfile should reference the yarn registry as CI can't resolve this.

packages/styletron-engine-snapshot/README.md
```diff
@@ -0,0 +1,40 @@
...
+yarn add --dev styletron-engine-snapshot
+```
+
+## Reasoning
+
+Snapshots generated with `styletron-engine-atomic` have 2 main issues
+
+1. non-debuggable classNames in snapshots (`"ae af "`)
+
+2. className generation is non-deterministic. If you create a new component,
```
rtsao:
```suggestion
2. className generation is dependent on internal engine state. If you create a new component,
```

packages/styletron-engine-snapshot/src/create-render.js
rtsao: I suppose we might as well just remove this comment entirely if documented in the README.

packages/styletron-engine-snapshot/src/server/server.js
```diff
@@ -0,0 +1,33 @@
+// @flow
+/* eslint-env browser */
```
rtsao:
```suggestion
```

packages/styletron-engine-snapshot/src/client/client.js
```diff
@@ -0,0 +1,33 @@
+// @flow
+/* eslint-env browser */
```
rtsao:
```suggestion
```

packages/styletron-engine-snapshot/src/index.js
```diff
@@ -0,0 +1,4 @@
+// @flow
+
+export {default as Client} from "./client/client.js";
```
rtsao: Given that the client and server implementation are currently identical, I think this could just export the same module under both `Client` and `Server` aliases.

packages/styletron-engine-snapshot/src/create-render.js
rtsao: Since this shouldn't be used in an actual browser (as no styling works), it might be worth considering using `util.inspect`:

```js
const style = {
  zzz: "zzz"
  ":after": {
    content: `"Hello world"`
  },
  aaa: "aaa"
};

util.inspect(style, {depth: Infinity, sorted: true});
// -> "{ ':after': { content: '\"Hello world\"' },\n  aaa: 'aaa',\n  zzz: 'zzz' }"
```

packages/styletron-engine-snapshot/src/create-render.js
koulmomo: I've severely stripped down the comment to just the essence of what this should output

packages/styletron-engine-snapshot/src/create-render.js
koulmomo: In this case I adopted an approach that makes use of [`prettier`](https://prettier.io/), which uses an `AST` to "pretty print" code.

**Why prettier instead of using the [suggested `util.inspect`](https://github.com/styletron/styletron/pull/342#discussion_r335203582)?**
When exploring using `util.inspect` I ran into a number of issues:

1. Using pure `util.inspect` would require adding a strict node `engine` requirement to `>10` on the package in order to support [`sort` option](https://nodejs.org/docs/latest-v10.x/api/util.html#util_util_inspect_object_options)

2. For backwards compat (to Node v8), one option would be to pre-sort the object before passing it to `util.inspect`.
Packages accomplish this usually by recursively recreating the object in alphabetical order, therefore making sure that the insertion order is
alphabetical.

ie
```js
// illustrative purposes only,
// actual implementation would need to recursively sort nested objects
// + any objects inside of arrays, as well as do some sanitization logic
// before calling `Object.keys`
function sortObj(obj) {
  var sortedObj = {};
  Object.keys(obj).sort().forEach(key => {
    sortedObj[key] = obj[key];
  });
  return sortedObj;
}
```
However, this is [not guaranteed by spec](https://stackoverflow.com/a/30919039), though [in practice this shouldn't be an issue in v8](https://github.com/nodejs/node/blob/master/lib/internal/util/inspect.js#L413)

Was originally going to continue with this approach until realizing that pre `v10`, `util.inspect` does not support the `compact` option, which is needed for `util.inspect` to return a properly indented + newlined string.

**Prettier approach**
The prettier approach instead [works in the browser + node](https://prettier.io/docs/en/browser.html).

The idea is to generate a string that is setting a variable equal to the result of sorted `JSON.stringify`,
and then running that result through `prettier` with a number of options to pretty-print the declaration and strip unnecessary quotes.

example:

In
```js
style = {":after": {"content": "\"Hello World\""}, "color": "red", "zIndex": "-1000"}
```

Out
```js
style = {
  ':after': {
    content: '"Hello World"'
  },
  color: 'red',
  zIndex: '-1000'
}
```

packages/styletron-engine-snapshot/src/index.js
koulmomo: done

packages/styletron-engine-snapshot/src/create-render.js
```diff
@@ -0,0 +1,46 @@
...
+        // don't add a semicolon at the end of the declaration
+        semi: false,
+
+        // default/prefer use of `'` for strings when needed
+        singleQuote: true,
+
+        // force newline insertions more often
+        printWidth: 40,
+      })
+      .replace(`${kind} = `, `${kind}=`);
```
rtsao: Rather than this replacement, it might be preferable to use the [prettier json5 parser](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEwA6yCGAzeAnDJAAnXGnlkKIzVpAAk4AbJiIgdQjyYBNaaQAXwA01MqwLIxeOHxCiMALwCSUHnAAeVDAFoAjAAYjGQfJAQADjACW0AM7JQmPHggB3AArOEDlJgBuENY8ZgBGeJhgANZwMADKFpHWUADmyNiYTHZwwiAAFjAAtkzsedbwdolgcHE+5db+5QCeyOB2DrnJ2XgwHhEphZjpmdm5AFZ2GgBCEdGxcZiFcAAyyXDDWTkgExpxySlMcACKAK4Q8BujIIl43a0T0ACsZhZ4yTDswTB5yAAsBrlXhBsuwIhZWpk3Jgmh0QABHM7wPqWXwgTB2HRQOCyWRmGQI6wyPqYAZDJAZTa5bKFazIGB4E5bOz7Q6nc7rJD0xm5GCYUKfHjfZAAJh5EWsTH2AGEIIVBq0oNB1rkTtkACp83wU7KCQRAA) and then concatenate whatever prefix you want

koulmomo: ooh nice.

Using the `json5` parser + string prefixing is definitely more legible and less `hacky`. It's probably also more performant/faster since it's a much simpler grammar than full fledged js.

Unfortunately it turns out that there is no standalone `json5` parser and it is instead [bundled s part of `babylon`](https://github.com/prettier/prettier/blob/a093bb3f7b9f59d8cbaf7e20f97f6fafceaef21b/src/common/internal-plugins.js#L28), so we wont see the build get smaller 🙁

rtsao: Ah, oh well 🤷‍♂

PR #343, opened 2019-10-18
docs(contrib): add guide and mini app

gergelyke:


packages/playground/next.config.js
```diff
@@ -0,0 +1,9 @@
+/* eslint-env node */
+
+module.exports = {
+  webpack: function(config) {
+    config.externals = config.externals || {};
+    config.externals["styletron-server"] = "styletron-server";
```
rtsao: I think this is outdated. There's no `styletron-server` package anymore

PR #351, opened 2019-11-08
Replace usage of dataset with getAttribute

rtsao:
IE 10 doesn't support `dataset`: https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes#Issues

So all else being equal, using `getAttribute` seems preferable for wider compatibility.


rtsao:
> do we support / try to support IE10?

No, and I don't think we should spend any extra effort here. That said, this is a simple change that will prevent at least one polyfill from being needed if folks did care about IE10.

Issue #341, opened 2019-10-03
Feature request: Export StyleObject in styletron-react


phou87:
We use a lot of custom styles in our codebase, and as a result end up passing style objects to components (primarily BaseUI components) a lot. Right now, we type these objects with our own custom type, but it'd be nice to be able to use StyleObject instead.

We could import it from styletron-standard, but we'd ideally not want to add a direct dependency on that package since it would be literally the only thing we're using.

chasestarr:
Could you provide an example? Does the pattern below not infer the type as-expected? 

```js
const customStyle = {
  color: 'blue',
  fontSize: '10px',
};

const A = styled('div', customStyle);
```

phou87:
It's moreso when we're making our own components customizable via style props.

```
import {Button} from 'baseui/button';

type Props = {
  style?: StyleObject,
};

function MyComponent(props) {
  return (
    <Button
      style={{...props.style, marginLeft: '16px'}}
    />
  );
}
```

chasestarr:
I generally don't recommend allowing arbitrary styles provided to a component and initially defer towards defining a more restrained interface. Doing so would make it easier to edit the internal details of your component definition without the risk of adversely affecting instances of it.

Recommendations aside, I believe this is a fine addition to the `styletron-react` package. It already relies on the `StyleObject` in it's public interface and any breaking change introduced to `styletron-standard` requires a major bump in `styletron-react` as well.

PR #353, opened 2019-11-22
Export StyleObject type from styletron-react package

chasestarr:
closes https://github.com/styletron/styletron/issues/341

### DIFF
--- a/packages/styletron-react/src/index.js
+++ b/packages/styletron-react/src/index.js
@@ -32,7 +32,7 @@ import {
 } from "./dev-tool.js";
 
 export {DebugEngine};
-export type StyleObject = StdStyleObject;
+export type {StyleObject};
 
 const noopEngine = {
   renderStyle: () => "",
--- a/packages/styletron-react/src/index.js
+++ b/packages/styletron-react/src/index.js
@@ -11,7 +11,7 @@ import {
   driver,
   getInitialStyle,
   type StandardEngine,
-  type StyleObject as StdStyleObject,
+  type StyleObject,
 } from "styletron-standard";
 
 import type {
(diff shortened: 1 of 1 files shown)


rtsao:
I think there might need to be some additional changes to the style types (for newer versions of Flow) so might want to hold off exporting them until that is more settled.

phou87:
How are we looking on this?

phou87:
Ping - given that this style is already exported, just in styletron and not styletron-react, any changes will need to be resolved anyways (anyone wishing to import the type currently will just do it from the former package), so I don't see the downside of exporting this now to save people a direct dependency?

chasestarr:
Given that this PR improves type ergonomics and does not solve a bug nor add a feature we're going to hold off and make a single change once everything is ready to publish

phou87:
Ok, is there an ETA for this?

packages/styletron-react/src/index.js
```diff
@@ -11,7 +11,7 @@ import {
   driver,
   getInitialStyle,
   type StandardEngine,
-  type StyleObject,
+  type StyleObject as StdStyleObject,
```
rtsao: Won't this break existing usages of `StyleObject`?
```suggestion
  type StyleObject,
```

packages/styletron-react/src/index.js
```diff
@@ -32,6 +32,7 @@ import {
 } from "./dev-tool.js";
 
 export {DebugEngine};
+export type StyleObject = StdStyleObject;
```
rtsao:
```suggestion
export type {StyleObject};
```

packages/styletron-react/src/index.js
rtsao: Nevermind, type declarations are hoisted. But I still think we can just export the actual name

Issue #21, opened 2016-12-06
Add support for css feature queries


stevenbenisek:
[Feature Queries](https://developer.mozilla.org/en/docs/Web/CSS/@supports) seem doable without adding to much extra code. 
@rtsao I could create a PR if this is something you think Styletron could benefit from.

rtsao:
I've actually never really used feature queries, but this seems pretty interesting.

I wonder if this could somehow help solve https://github.com/rtsao/styletron/issues/6 where there's reusable atomic "fallbacks" based on feature queries.

Are there any other CSS-in-JS libraries that support feature queries? I'd be interested in seeing how they are used and what the syntax is like.

stevenbenisek:
Both [cxs](https://github.com/jxnblk/cxs) and [Free-Style](https://github.com/blakeembrey/free-style) support feature queries. The syntax is equal to that of media queries. In the case of Styletron it would look like this:

```js
import {injectStyle} from 'styletron-utils';

injectStyle(styletron, {
  '@supports (flex-wrap: wrap)': {
      display: 'flex',
      flexWrap: 'wrap'
  }
});
```
This is an excellent way to progressively enhance the experience. 
e.g: Android < 4.4 has support for `display: flex` but not for `flex-wrap: wrap`. Just applying both properties might break the layout of the component. Using feature queries to test if `flex-wrap` is supported mitigates this problem.

Feature queries could be used to provide value fallbacks as per #6. Free-Style has an interesting way of soving this: [https://github.com/blakeembrey/free-style#overload-css-properties](https://github.com/blakeembrey/free-style#overload-css-properties)

leyanlo:
Would also love to have this to be able to conditionally style padding around the iPhone X notch. E.g.
```css
@supports(padding: max(0px)) {
    .post {
        padding-left: max(12px, constant(safe-area-inset-left));
        padding-right: max(12px, constant(safe-area-inset-right));
    }
}
```
from https://css-tricks.com/designing-websites-iphone-x/

rtsao:
I think supporting `@supports` this makes sense. That said, I think there are a few questions that need to be resolved, mostly around specificity/precedence.

For example, what should the text color be in the case of the style below?
```js
const style = {
  "@supports (color: green)": {
    color: "green"
  },
  "@supports (color: blue)": {
    color: "blue"
  },
  color: "red",
}
```



Unlike media queries, there doesn't really seem to be an obvious way to sort these.

[...]

PR #356, opened 2020-01-22
Add styletron-engine-monolithic

tajo:
This is an implementation of styletron-standard that uses similar strategies as [emotion](https://emotion.sh/docs/introduction). It even uses parts of emotion (hashing and sheets handling). It should work as a drop-in alternative to styletron-engine-atomic. It enables some extra features with a different set of trade-offs:

- classNames are stable (hashes) so components can be cross-referenced 
- you can mix longhand and shorthand properties
- no need for media query sorting, since the order is always respected
- supports `@supports`

**TODOs (in following PRs):**
- fallback support (but maybe `@support` is a better alternative?)
- debug mode creates a lot of style elements with sourcemaps but that's probably ok
- documentation

The implementation is much simpler than atomic. Each style object is hashed and that hash is used as the className and cache key. We still keep three separate caches (keyframes, fonts and rest) so we can enforce the initial order (important for keyframes and fonts). 

On the server, we create a single `<style />` element. The attribute `data-hydrate` gets the list of all hashes. We parse them on the client and hydrate the cache there. We don't even need the values, just the information that class is already defined (coming from the server) is enough.

On the client, we use the sheet implementation from emotion. In the slow (dev) mode we create a style tag for each new class. In the fast (prod) mode, we group classes together and use `sheet.insertRule`.

I already published first `0.0.1` version just to test it. We can keep it as experimental for now and once tested in a few apps, we can bump to v1.

[Base Web running monolithic engine](https://baseweb-git-monolithic.uber-ui-platform.now.sh).


leyanlo:
> * classNames are stable (hashes) so components can be cross-referenced
Does this mean Styletron would support descendant hover/focus, too?? :-D

> * you can mix longhand and shorthand properties
Does this mean supporting tagged template literals? Otherwise how would we guarantee order of keys in a JS object?

So excited for this!

tajo:
> > * classNames are stable (hashes) so components can be cross-referenced
> 
> Does this mean Styletron would support descendant hover/focus, too?? :-D

Yes, that should work now. Need to test it more tho.

> > * you can mix longhand and shorthand properties
> 
> Does this mean supporting tagged template literals? Otherwise how would we guarantee order of keys in a JS object?

Well, I know the order is not guaranteed by a spec but it is preserved by all modern JS engines so it's pretty safe to rely on that.

Tagged template literals are not currently supported (and not planned for now).

rtsao:
Nice, this will be great!

I think the style object API ought to be the same between the monolithic and atomic engines so components remain portable across engines. At the very least, if there are any differences, I think the monolithic engine should have a "portability mode" that, in development, provides warnings that indicate patterns that would not work in an atomic engine, so it's clear whether the component is portable or not.

As far as descendant combinators, I think that's a somewhat orthogonal issue to atomic vs monolithic rendering. Although it is certainly easier to implement with hashed, content-based generation of class names, the root issue in my view is really just the nature of style objects in conjunction with the "selectorless", class-free paradigm of `styled`.

That said, I think descendant combinators definitely make a lot of sense with the `useStyletron` API, because fundamentally you are composing the hierarchy of elements and classnames directly within a single component, which I think sidesteps the issues of heirarchy-dependent selectors with the `styled` API components.

So I think that using `useStyletron`, we can make an API for descendant combinators that works using both engines.

tajo:
> I think the style object API ought to be the same between the monolithic and atomic engines so components remain portable across engines. 

Not really possible (or practical) with features like `@supports`, combinators relying on stable classNames or mixing longhands / shorthands. Those are all things that many developers want to do and our major CSAT killer. 

I think it's fine to accept that monolithic is a superset of atomic in terms of features. It just needs a  good documentation to explain the trade-offs.

> At the very least, if there are any differences, I think the monolithic engine should have a "portability mode" that, in development, provides warnings that indicate patterns that would not work in an atomic engine, so it's clear whether the component is portable or not.

We could duplicate some of the validation code that's in atomic but changing some "portability" flag is not really easier than swapping engines (it takes to replace a single import). I would rather avoid the duplication.

> So I think that using `useStyletron`, we can make an API for descendant combinators that works using both engines.

Monolithic engine is already more permissive, so you can do this right now:

```jsx
const [css] = useStyletron();
<div
    className={css({
      padding: "100px",
      border: "1px solid black",
      ":hover > div": {backgroundColor: "yellow"},
    })}
  >
  <div className={css({padding: "30px", border: "1px solid blue"})}>
    This background changes when parent is hovered
  </div>
</div>
```

Also, since `css` returns a stable className, you can also use those to define selectors:

```jsx
const childCx = css({padding: "30px", border: "1px solid blue"});
const parentCx = css({
  padding: "100px",
  border: "1px solid black",
  [`:hover > .${childCx}`]: {backgroundColor: "yellow"},
});

<div className={parentCx}>
  <div className={childCx}>
    This background changes when parent is hovered
  </div>
</div>
```
(the extra debug className needs to be removed from childCx but it works). 

I think this already covers the most wanted use-case and there is no need to invent a new API. Also it uses just css. The first example works with `styled()` too. 

Personally, I'm still happy to use atomic CSS in all my projects and avoid these patterns altogether. I think the atomic restrictions make sense and help to build better components in a long-term. Also pretty sure, that Base Web will keep using atomic so it will be always compatible with both.

rtsao:
I believe `@supports` is also an issue completely orthogonal to atomic vs monolithic. This could be implemented in the atomic engine but the [questions I had](https://github.com/styletron/styletron/issues/21#issuecomment-481498288) are related to ambiguity of the semantics of style objects in general, which equally apply to the monolithic engine.

Agreed that mixing shorthand longhand should be allowed in the monolithic engine, but I wouldn't really consider that an API difference (and something that be checked in a "strict/portability mode" of the monolithic engine).

So I think having the same styling API between the atomic and monolithic engines is both worthwhile and not really a barrier to anything.

Regarding child selectors:

```js
const [css] = useStyletron();
<div
    className={css({
      padding: "100px",
      border: "1px solid black",
      ":hover > div": {backgroundColor: "yellow"},
    })}
  >
  <div className={css({padding: "30px", border: "1px solid blue"})}>
    This background changes when parent is hovered
  </div>
</div>
```

Even this works today with the atomic engine, it's just discouraged since it relies on global selectors.

Of course, your second example illustrates why content-based hashing makes it easier for descendent combinators. The second example won't work today with the atomic engine. That said, I think there's ways we can implement this in the atomic engine, even with the API you propose. So I definitely think we should add descendant combinators (at the very least into `useStyletron`), I just think it should work for both engines, so the component definition APIs remain agnostic of the engine and therefore components remain portable.

tajo:
Let's keep the discussion about `@supports` in #21. The rest makes sense to me.

> That said, I think there's ways we can implement this in the atomic engine, even with the API you propose. So I definitely think we should add descendant combinators (at the very least into useStyletron)

Do you have something specific in mind to preserve this API?

```jsx
const childCx = css({padding: "30px", border: "1px solid blue"});
const parentCx = css({
  padding: "100px",
  border: "1px solid black",
  [`:hover > .${childCx}`]: {backgroundColor: "yellow"},
});

<div className={parentCx}>
  <div className={childCx}>
    This background changes when parent is hovered
  </div>
</div>
```

My idea:
1. You could create the same hashed className out of child styles (although this would sadly add an overhead to atomic engine or styletron-react)
2. Put it to the first position of `childCx`, so it would look like `css-kjh123 ae ab cb`
3. `css()` would "secretly" autofix the selector from `:hover > css-kjh123 ae ab cb` to `:hover > .css-kjh123`.

I guess you somehow need to generate a stable selector for the child component in order to support that API. Or you could add a second parameter to `css()` and let the user to define it but that might be a bit annoying (especially when not necessary for monolithic).

tajo:
@rtsao 

How about deprecating `styletron-engine-atomic`? **Some significant pros:**

1. No need trying to backport features from monolithic.
2. Less confusing for users. It's hard to explain nuanced trade-offs.
3. Easier to maintain since cutting the codebase significantly. We also don't have to roll out two plugins for fusion, gatsby... or maintain two sets of documentation. 

The main (and only?) reason why would we want to use atomic over monolithic is performance so I did a quick benchmark. I used [the one](https://github.com/tajo/emotion/tree/master/scripts/benchmarks/src/implementations) from emotion since it's up to date and was easy to spin up. (I saw [yours](https://github.com/rtsao/css-in-js-perf-tests) that looks very comprehensive but it errors out for me and would probably take some effort to update since it's quite old).

**Results:**

| | emotion-styled | styletron-atomic | styletron-monolithic |
| --- | --- | --- | --- |
| Mount deep tree            | 15.18 ±00.41 ms | 18.11 ±00.76 ms | 16.85 ±00.46 ms |
| Mount wide tree             | 19.24 ±00.79 ms | 20.00 ±02.88 ms | 20.41 ±03.35 ms |
| Update dynamic styles | 16.80 ±02.46 ms | 11.05 ±01.06 ms | 14.55 ±01.01 ms |

All three examples use `styled` function (emotion has almost an identical API). **It seems there is almost no difference at all in runtime performance**.

How about bundle sizes? I compared the button page from baseweb.design (heavy use of styles) and homepage of eats (lighter use of styles).

```
116708    baseweb-atomic.html
14074     baseweb-atomic.html.br
191069    baseweb-mono.html
17776     baseweb-mono.html.br
59169     eats-atomic.html
11561     eats-atomic.html.br
65050     eats-mono.html
12086     eats-mono.html.br
```

I built it with both engines and also did the compression with brotli (should be more important number for real scenarios.) **Again, the difference is pretty small (brotli) - ~4kB for Base Web and ~0.5kB for eats**.  To put it into perspective, both apps load more than 1MB in other resources.

Atomic engine is more efficient when it comes to bundle sizes but it might not be worth it? The most of CSS in JS ecosystem (emotion and styled-components) happily uses the monolithic approach and it's definitely more developer friendly.

packages/styletron-engine-monolithic/src/server/server.js
```diff
@@ -0,0 +1,135 @@
...
+      );
+    }
+    return `${this.opts.prefix || ""}font-${fontName}`;
+  }
+
+  renderKeyframes(keyframes: KeyframesObject): string {
+    const animationName = hashCssObject(keyframes);
+    if (!this.keyframesCache[animationName]) {
+      this.keyframesCache[animationName] = keyframesBlockToRule(
+        `${this.opts.prefix || ""}animation-${animationName}`,
```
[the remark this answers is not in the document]
tajo: I used the same naming as emotion since I liked the readability. 

Brotli should take care of it - it even utilizes a dictionary with some common words where you can find `font` or `ation`. Also, keyframes and font-faces are relatively rare compare to other css rules.

hc-codersatlas:
@tajo the benchmarks doesn't consider memory usage, which would likely favor atomic. I also thought the idea of atomic was being able to reuse styles? Most of these single component benchmarks doesn't cater for this and so shows atomic at a disadvantage.

tajo:
> @tajo the benchmarks doesn't consider memory usage, which would likely favor atomic

@hc-codersatlas `CSSStyleSheet` will have more CSS rules for sure. On the other hand, the internal state of monolithic engine takes less memory since the cache is a simple map of `className <=> boolean` where atomic cache needs to map `style object <=> className`.

I think the browser memory is not a big concern compared to the runtime performance and network data. Do you have some specific benchmark you want to run?

packages/styletron-engine-monolithic/src/client/client.js
```diff
@@ -0,0 +1,142 @@
...
+      // there is a single style tag coming from the monolithic server
+      const element = opts.hydrate[0];
+      const dataHydrate = element.getAttribute("data-hydrate");
+      if (dataHydrate) {
+        dataHydrate.split(" ").forEach(hashKey => {
+          // cache keys are unique across fonts, keyframes and other css so
+          // we use a single attribute to get them all
+          this.styleCache[hashKey] = true;
+          this.fontFaceCache[hashKey] = true;
+          this.keyframesCache[hashKey] = true;
```
rtsao: If we hydrate in this manner, I wonder if it'd be preferable to just share the same cache reference for all three aliases.

Since I think the only benefit of separate caches would be smaller size and thus faster lookups, but if they contain the same keys this benefit seems greatly reduced.

packages/styletron-engine-monolithic/src/client/client.js
```diff
@@ -0,0 +1,142 @@
...
+  fontFaceSheet: HTMLStyleElement;
+  keyframesSheet: HTMLStyleElement;
+
+  styleSheet: StyleSheet;
+  styleCache: cacheT;
+  keyframesCache: cacheT;
+  fontFaceCache: cacheT;
+
+  constructor(opts?: optionsT = {}) {
+    this.opts = opts || {};
```
rtsao: I still think it'd be nice to have a "strict/compat" mode that in development logs the same warnings as the atomic engine.

packages/styletron-engine-monolithic/src/client/sheet.js
```diff
@@ -0,0 +1,130 @@
...
+
+    if (this.isSpeedy) {
+      const sheet = sheetForTag(tag);
+      rules.forEach(rule => {
+        try {
+          // this is a really hot path
+          // we check the second character first because having "i"
+          // as the second character will happen less often than
+          // having "@" as the first character
+          const isImportRule =
```
rtsao: Should we even support `@import` rules?

packages/styletron-engine-monolithic/src/inject-style-prefixed.js
```diff
@@ -0,0 +1,98 @@
...
+      plainBlock += `${propValPair};`;
+    } else if (originalKey[0] === ":") {
+      outputRules.push(
+        `.${globalPrefix}css-${className}${originalKey}{${injectStylePrefixed(
+          originalVal,
+          "",
+          globalPrefix,
+        ).join("")}}`,
+      );
+    } else if (originalKey[0] === "@") {
```
rtsao: I think we should consider aligning the `@media` ordering strategy between the atomic and monolithic engine so that style object semantics remain as close as possible.

It seems like there's two options:
1. Add sorting to the monolithic engine. As we discussed though, this is less flexible for developers.
2. Replace sorting in the atomic engine with iteration order-based precedence. This could potentially be an option that is disabled by default to avoid a breaking change.

hc-codersatlas:
@tajo my experience is that memory matters. Maybe often overlooked. Most crashes I've seen (or freezes) aren't from CPU but memory usage. There's the garbage collection cost to pay for.

In terms of performance, shouldn't a group of atomic classes be cached?

e.g. there's a button and that button is likely to have the same set of atomic classes and be reused everywhere. For components with a lot of styles this would benefit a lot and is likely why monolithic wins.

packages/styletron-engine-monolithic/src/client/client.js
```diff
@@ -0,0 +1,142 @@
...
+    this.styleCache = {};
+    this.fontFaceCache = {};
+    this.keyframesCache = {};
```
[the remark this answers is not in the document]
tajo: Ah, I thought that `@keyframe` and `@font-face` needs to be defined before being used so this was a way to enforce the correct order but that's not the case so merging into a single cache object.

packages/styletron-engine-monolithic/src/client/sheet.js
tajo: removed

packages/styletron-engine-monolithic/src/inject-style-prefixed.js
tajo: In that case, I think number 2 makes sense:

1. We don't need to add this arbitrary limitation into monolithic engine
2. It will make the atomic engine more flexible
3. The implementation could share the logic with planning `@supports` support

tajo:
It seems that `unitest@2.1.2` breaks unit tests, `2.1.2` works fine. We made this update with @rtsao to update `chrome-launcher` since the older version doesn't work on `macOS Catalina`. I was not able to pinpoint the root cause so let's keep the current `unitest` in place but we should look into it so macOS developers can run tests locally.

gergelyke:
@tajo @rtsao is there anything else we need to do before landing it?

tajo:
@gergelyke I think all the feedback should be resolved so waiting for final ✅by @rtsao (not sure if he wants to update the atomic engine first to have functional parity).

PR #369, opened 2020-07-17
Update cup3

zxbodya:
- update create-universal-package to the latest version in v3 branch
- fix an issue with never version of babel complaining about duplicated variables


tajo:
we will do cupv4

PR #370, opened 2020-07-20
removes flow message tests

chasestarr:
Likely want to restore this type of test after TS


tajo:
I am going to delete this package as a part of the typescript PR.

PR #371, opened 2020-07-21
[WIP] Typescript migration

tajo:
This PR is based on @zxbodya's work. It adds new build process for **styletron-react** that uses this version of cup: https://github.com/rtsao/create-universal-package/pull/47 (published as `@miksu/create-universal-package` for now). 

It also uses tsc, api-extractor and flowgen to extract both ts and flow libdefs (part of the build). The resulting flow libdefs have some anys and are less strict. On the other hand, ts libdefs are better than what we currently have on definitely-typed. 

I published this as `@miksu/styletron-react` and tested it out in baseui. It works!

TODO:
- copy the build process to all packages
- figure out what to do about testing, probably using jest and puppeteer, @chasestarr is also looking into that


packages/styletron-react/package.json
```diff
@@ -12,7 +12,7 @@
   ],
   "main": "./dist/index.js",
   "module": "./dist/index.es.js",
-  "types": "./lib/index.d.ts",
```
zxbodya: this might is a bit tricky one - for editing experience and in particular for project references to work correctly, it might need to point to lib with compilation results

packages/styletron-react/src/types.ts
```diff
@@ -0,0 +1,96 @@
...
+  ref: Props extends {ref?: infer T} ? T : React.Ref<any>;
+}>;
+
+export type StyletronComponent<Props> = React.FC<
+  Props & StyletronProps<Props>
+> & {
+  __STYLETRON__: any;
+};
+
+export type StyledFn = {
```
chasestarr, line 60: I wonder if typescript's function overloading would be more appropriate https://www.typescriptlang.org/docs/handbook/declaration-files/by-example.html#overloaded-functions and https://www.typescriptlang.org/docs/handbook/functions.html#overloads

packages/styletron-standard/package.json
```diff
@@ -30,7 +31,7 @@
     "prepublish": "npm run build"
   },
   "dependencies": {
-    "@rtsao/csstype": "2.6.5-forked.0",
+    "csstype": "2.6.9",
```
chasestarr, line 34: This may create some cryptic flow error messages and was why we added the flow error message tests

packages/styletron-standard/src/index.ts
```diff
@@ -8,19 +6,10 @@ import type {
...
-//
-//
-export type StyleObject = $Shape<{
-  ...Properties,
-  [string]: StyleObject, // Unrecognized properties are assumed to be media queries or pseudo selectors w/ nested style object. See: https://github.com/styletron/styletron-standard
-}>;
+export interface NestedStyleObject {
+  [x: string]: StyleObject;
+}
+export type StyleObject = NestedStyleObject | Properties;
```
chasestarr, line 12: This feels inaccurate, but maybe it's my unfamiliarity with TS. Can I write a style object like below? It looks to me that it would not allow combining properties + nested properties

```
{
  backgroundColor: 'blue',
  ':hover': {
    backgroundColor: 'red',
  }
}
```

packages/styletron-standard/src/style-types.ts
```diff
@@ -0,0 +1,43 @@
...
+  StandardProperties,
+  VendorProperties,
+  ObsoleteProperties,
+  SvgProperties,
+  AnimationNameProperty as CTAnimationNameProperty,
+  FontFamilyProperty as CTFontFamilyProperty,
+  FontFace as CTFontFace,
+} from "csstype";
+
+export interface KeyframesPercentageObject {
```
chasestarr, line 11: should this be exported?

packages/styletron-react/src/__tests__/tests.browser.tsx
```diff
@@ -13,9 +11,10 @@ import {
   withTransform,
   Provider,
   useStyletron,
-} from "../index.js";
+} from "../index";
 
 import {getInitialStyle, driver} from "styletron-standard";
+import {StyleObject} from "styletron-standard";
```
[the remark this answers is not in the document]
zxbodya: this probably should be `import type` instead

packages/styletron-engine-atomic/src/css.ts
```diff
@@ -13,7 +11,7 @@ export function atomicSelector(id: string, pseudo: string): string {
   return selector;
 }
 
-export function keyframesToBlock(keyframes: {[string]: Object}): string {
+export function keyframesToBlock(keyframes: {[x: string]: any}): string {
```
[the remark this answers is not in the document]
zxbodya: probably, was not really focusing on improving types(this is exactly what it was in flow types)

I think it might better to have type improvements as separate from this, as an option starting from enabling strict mode which will highlight where to start :)

package.json
```diff
@@ -7,25 +7,29 @@
...
   },
   "devDependencies": {
+    "@microsoft/api-extractor": "^7.9.2",
+    "@typescript-eslint/eslint-plugin": "^3.6.1",
+    "@typescript-eslint/parser": "^3.6.1",
     "babel-eslint": "^10.0.3",
     "eslint": "^6.3.0",
     "eslint-config-cup-recommended": "^2.0.2",
     "eslint-plugin-cup": "^2.0.1",
     "eslint-plugin-flowtype": "^4.3.0",
```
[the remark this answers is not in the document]
zxbodya: kept it for use to run existing flow tests with flow types generated from typescript, to better see the difference

packages/styletron-react/src/index.tsx
```diff
@@ -24,15 +27,33 @@ import type {
...
+  NonAssignmentCommutativeReducerContainer,
+  StyledFn,
+  WithStyleFn,
+  WithTransformFn,
+  WithWrapperFn,
+  StyletronProps,
+  addDebugMetadata,
+  setupDevtoolsExtension,
+  DebugEngine,
+};
```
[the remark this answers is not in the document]
zxbodya: this is not strictly required but recommended by api-extractor (it will show warning for types used in exports but not exported)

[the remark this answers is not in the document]
chasestarr: My concern is that changes to largely internal details will require major version releases

packages/styletron-engine-atomic/src/sort-css-media-queries.ts
```diff
@@ -27,7 +25,7 @@ function _getQueryLength(length: string) {
   if (matches === null) {
     return maxValue;
   }
-  let number = matches[1];
+  let number: number | string = matches[1];
```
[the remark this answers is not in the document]
zxbodya: it is reassigned later in code… generally - I would prefer to split it into two separate variables matched string and resulting number

packages/styletron-engine-atomic/src/css.ts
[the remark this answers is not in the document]
chasestarr: Ah I didn't realize that this fellback to `any` in flow. Thought it would at least loosely check https://flow.org/try/#0C4TwDgpgBA8lC8sBGArCBjYBuAUD9A9gHYDOwUAhgFywJQCMWU+xZUSNciA5Bd7oVLl0nOkQCuAG0lYgA

package.json
```diff
@@ -7,25 +7,29 @@
...
     "bootstrap": "lerna bootstrap",
     "clean": "lerna clean",
-    "lint": "eslint --ignore-path .gitignore .",
-    "pretest": "yarn run lint && yarn run flow",
+    "lint": "eslint --ignore-path .gitignore --ext js,ts,tsx .",
+    "pretest": "yarn run lint && yarn run tscheck",
     "test": "lerna run test",
     "build": "lerna run build",
-    "flow": "flow --max-warnings=0",
+    "tscheck": "tsc -b",
```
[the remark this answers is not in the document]
tajo: afaik tsc doesn't really produce warnings? everything is an error, we can change the strictness (it's going to be low to make the transition easier)

packages/styletron-react/src/index.tsx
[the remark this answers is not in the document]
tajo: There are some good reasons why they ask developers to export these types: https://api-extractor.com/pages/messages/ae-forgotten-export/

packages/styletron-react/package.json
tajo: what do you mean? We are just keeping them in `types: "./dist/index.d.ts"`

packages/styletron-react/package.json
```diff
@@ -1,33 +1,31 @@
 {
   "name": "styletron-react",
-  "version": "5.2.7",
+  "version": "5.2.8",
   "description": "React bindings for Styletron",
   "author": "Ryan Tsao <ryan.j.tsao@gmail.com>",
   "repository": "styletron/styletron",
   "sideEffects": false,
   "files": [
     "dist",
```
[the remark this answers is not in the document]
tajo: `dist` now contains the bundled version of ts libdefs (output of api-extractor)

packages/styletron-react/src/index.tsx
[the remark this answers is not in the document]
chasestarr: Has this been requested by flow users? I would prefer to take a less open approach and add the feature if requested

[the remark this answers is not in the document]
zxbodya: > My concern is that changes to largely internal details will require major version releases

but it does not really change… it requires type to be exported only if it is already exported as part of something, so technically speaking it is already part of public api

however big part of features of api extractor is actually about managing scope of what should be public api, instead we can add `@internal` in jsdoc for some methods to avoid publishing them

api-extractor.json
```diff
@@ -0,0 +1,27 @@
...
+    "enabled": true,
+    "untrimmedFilePath": "<projectFolder>/dist/index.d.ts"
+  },
+  "tsdocMetadata": {
+    "enabled": false
+  },
+  "newlineKind": "lf",
+  "messages": {
+    "extractorMessageReporting": {
+      "ae-missing-release-tag": {
```
zxbodya, line 22: probably it was bad idea, when I disabled that rule… instead of adding multiple exports - it should be better to just declare internal things as `@internal` in jsdoc

packages/styletron-react/package.json
zxbodya: the difference would be only when changing something in referenced module, for example when editing `styletron-standard` with `types: lib/index` that definition file would be automatically recompiled when doing type check in `styletron-react` or other project depending on it… 

but the bundle in dist folder would not be recompiled automatically by typescript

packages/styletron-standard/package.json
chasestarr: Sorry, we had attempted to update it a few months ago but the newer version presented styletron users with hard-to-comprehend error messages. Just pointing out that we may find issues in this area

packages/styletron-react/src/index.tsx
```diff
@@ -154,7 +183,7 @@ export function useStyletron() {
   const hydrating = React.useContext(HydrationContext);
   checkNoopEngine(styletronEngine);
 
-  const debugClassName = React.useRef("");
+  const debugClassName = React.useRef<string | void>("");
```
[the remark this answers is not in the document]
chasestarr: missing flow inference already 😆

packages/styletron-standard/src/index.ts
zxbodya: actually, what is currently in a code looks to work the same way as in comment, and should be good for the use case we have: [playground with example](https://www.typescriptlang.org/play/index.html#code/C4TwDgpgBACgTge0nYBLCBnKBeWBDFVPAGwB4BvKAKClqgGMFiE4AuKDYOVAOwHMA3DToAjPPQDWfRAFceAEwDCTFu07d+QgL4A+KlV7AIcAGbjoAOUxH5AZVDEIAeREArCPWBRywgNoAPNS5ePgBddnsQRxd3T219UEgoSOi3Dy9cK04IOwdnNM8oAB9YRGQ0TCEqRh5OKCNOCLyY9JxvYTFJaQQ5JRU2KAAiEWIZCEGAGmFB1gALBAA3Yxn2uihOqVkFZWYBwbgcyeFaGfE0JZWfNdpGXfZ9w6m1rSeoF6otKpq6huAAJiaUXysQyqygAHpwVAAALADAAWgg-kgnkRcEQcA64k2PW2-XYAEY-gBmV4zeZLOCXY7rbHdXo7VRDA7yI5rU6eVAXdhXa6QmFwxHI9JojE0m74qAE160d6yqafBLWAB0txYQn5sIRSJRwFFLCov18AHI5otjMbQsqNvS8bsgA)

chasestarr: 👍 the playground example lgtm

api-extractor.json
tajo: We are not publishing the bundled version of types so this doesn't really matter?

packages/styletron-react/src/index.tsx
```diff
@@ -415,6 +452,7 @@ export function composeStatic(
         base: styletron.base,
         driver: styletron.driver,
         wrapper: styletron.wrapper,
+        // @ts-ignore
```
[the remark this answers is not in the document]
tajo: seems like conflicting types for array items, used any for the later part

api-extractor.json
tajo: (chatted about this more, we can figure out later how to better use api-extractor, for now we just use it to bundle types)

packages/styletron-standard/src/style-types.ts
```diff
@@ -0,0 +1,43 @@
...
+export type FontFamilyProperty = CTFontFamilyProperty | FontFace;
+
+export type TLength = string | 0;
+export type Properties = {
+  animationName?: AnimationNameProperty;
+  fontFamily?: FontFamilyProperty | FontFamilyProperty[];
+  MozAnimationName?: AnimationNameProperty;
+  WebkitAnimationName?: AnimationNameProperty;
+  OAnimationName?: AnimationNameProperty;
+} & Omit<
```
chasestarr, line 33: This `Omit` utility type is ending up in the flow type definitions generated. See published file here: https://unpkg.com/@miksu/styletron-standard@0.0.1/dist-node-cjs/index.js.flow

Another issue in the generated file is that import statements do not include the `type` prefix.

```
import { AnimationNameProperty as AnimationNameProperty_2 } from "csstype";
// should be
import type { AnimationNameProperty as AnimationNameProperty_2 } from "csstype";
```

zxbodya: I suspect flowgen is not yet supporting `import type` typescript feature (which was added relatively recently), same about `Omit` which is older but still probably new comparing to most of flowgen code

lifeiscontent:
@chasestarr why was this PR closed?

tajo:
> @chasestarr why was this PR closed?

We were trying to converts Styletron to TS and auto-generate flow types out of TS libdefs. Unfortunately, the later part is too hard so we are going to postpone the TS migration till we are ready to stop using flow completely in our projects.

PR #379, opened 2020-07-23
switch flow to ts, delete misnamed dir

chasestarr:


chasestarr:
interesting.. so we get the same test error in GH as well https://github.com/styletron/styletron/pull/379/checks?check_run_id=903647787#step:9:103

chasestarr:
turns out it was due to missing the class-properties babel transform

PR #380, opened 2020-07-23
Fix animation style definition

chasestarr:
switching from interface to type resolved lots of issues in the baseui codebase

### DIFF
--- a/packages/styletron-standard/src/style-types.ts
+++ b/packages/styletron-standard/src/style-types.ts
@@ -8,9 +8,9 @@ import type {
   FontFace as CTFontFace,
 } from "csstype";
 
-export type KeyframesPercentageObject = {
+export interface KeyframesPercentageObject {
   [key: string]: Properties;
-};
+}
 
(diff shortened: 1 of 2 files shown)


packages/styletron-standard/src/style-types.ts
```diff
@@ -8,9 +8,9 @@ import type {
   FontFace as CTFontFace,
 } from "csstype";
 
-export interface KeyframesPercentageObject {
+export type KeyframesPercentageObject = {
```
chasestarr, line 11:
```suggestion
export interface KeyframesPercentageObject {
```

packages/styletron-standard/src/style-types.ts
```diff
@@ -8,9 +8,9 @@ import type {
...
   [key: string]: Properties;
-}
+};
```
chasestarr, line 13:
```suggestion
}
```

zxbodya:
can you share example where this causes an issue? (my expectation is there should be no difference here when about using type alias or interface)

Issue #334, opened 2019-08-23
Streaming rendering support


rtsao:
This idea was initially proposed in https://github.com/styletron/styletron/issues/137 but I think there's a fairly good way of doing this.

**On the server:**
- Produce a stream of `<style>` tags from atomic engine
- Merge the `<style>` tag stream with the stream returned from `ReactDOM.renderToNodeStream`

**On the client:**
- Prior to `ReactDOM.hydrate`, hoist the contents of all the inline `<style>` elements into `<head>` and remove the inline `<style>` elements (while hydrating the cache as normal)

I think this would work flawlessly aside from the following (likely exceedingly rare) corner cases where interleaving of streams results in a `<style>` element being inserted:

A) as the first child in an element containing children styled with `:first-child`
B) as the last child in an element containing children styled with `:last-child`
C) into an element containing children styled with `:nth-child`
D) *between* two elements styled with `~` selectors

If this were to happen, then streaming rendering would cause incorrect styling (at least until the offending inline `<style>` is removed during hydration).

TxHawks:
I believe there are other cases where this could fail. Reliance on the cascade is indeed significantly decreased with atomic css, but it is not completely gone. 

For instance, the following could break until rehydration:
```jsx
// ComponentA.js
import { useStyletron } from "styletron-react";

export default () => {
  const [css] = useStyletron();
  return (
    <div
      className={css({ 
        backgroundColor: "#00f",
        "@media (min-width: 600px)": {
          backgroundColor: "#f00",
        },
      })}
    >
      I'm blue and turn red in screens wider than 600px
    </div>
  );
};
```

```jsx
// ComponentB.js
import { useStyletron } from "styletron-react";

export default () => {
  const [css] = useStyletron();
  return (
    <div
      className={css({ 
        backgroundColor: "#0f0",
        "@media (min-width: 600px)": {
          backgroundColor: "#f00",
        },
      })}
    >
      I'm green and turn red in screens wider than 600px
    </div>
  );
};
```

If because of streaming, server rendered html turns out to be:
```html
<style>.a{background-color: "#0f0"}</style>
<style media="(min-width: 600px)">.b{background-color: "#f00"}</style>
<div class="a b">I'm blue and turn red in screens wider than 600px</div>
<!-- ... -->
<style>.c{background-color: "#00f"}</style>
<div class="b c">I'm green and turn red in screens wider than 600px</div>
```

[...]

PR #394, opened 2021-03-16
add support for styles rendering and streaming

AyoubElk:
This closes [styletron/styletron/#334](https://github.com/styletron/styletron/issues/334)

#### Implementation details:
- Exposes a new method `getStylesheetsStream` that takes as an input a readable html stream and adds the newly rendered styles to it
- Clears older styles after each render
- Generates script tags that inject style tags in the document head
- New rendered styles are injected inside the appropriate media scoped style tag
- Style tags are ordered inside the head tag to support a mobile first approach

Looking forward to your feedback before adding tests/docs!
Thanks

#### Checklist
- [x] run tests
- [ ] tests updated
- [ ] documentation is changed or added

### DIFF
--- a/packages/styletron-engine-atomic/src/server/server.js
+++ b/packages/styletron-engine-atomic/src/server/server.js
@@ -130,7 +130,11 @@ class StyletronServer implements StandardEngine {
             },
           ]
         : []),
-      ...sheetify(this.styleRules, this.styleCache.getSortedCacheKeys()),
+      ...sheetify(
+        this.styleRules,
+        this.styleCache.getSortedCacheKeys(),
+        this.streamingMode,
+      ),
     ];
 
     // Clear streamed styles
@@ -257,7 +261,7 @@ function stringify(styleRules, sortedCacheKeys) {
   return result;
 }
 
-function sheetify(styleRules, sortedCacheKeys): Array<sheetT> {
+function sheetify(styleRules, sortedCacheKeys, streamingMode): Array<sheetT> {
   if (sortedCacheKeys.length === 0) {
     return [{css: "", attrs: {}}];
   }
@@ -265,8 +269,14 @@ function sheetify(styleRules, sortedCacheKeys): Array<sheetT> {
   sortedCacheKeys.forEach(cacheKey => {
     // omit media (cacheKey) attribute if empty
     const attrs = cacheKey === "" ? {} : {media: cacheKey};
+    let insertBeforeMedia =
+      streamingMode && styleRules[cacheKey].insertBeforeMedia
+        ? styleRules[cacheKey].insertBeforeMedia
+        : null;
+    if (streamingMode)
+      insertBeforeMedia = styleRules[cacheKey].insertBeforeMedia || null;
     sheets.push({
-      insertBeforeMedia: styleRules[cacheKey].insertBeforeMedia,
+      ...(insertBeforeMedia && {insertBeforeMedia}),
       css: styleRules[cacheKey].rules,
       attrs,
     });
--- a/packages/styletron-engine-atomic/src/server/server.js
+++ b/packages/styletron-engine-atomic/src/server/server.js
@@ -1,4 +1,5 @@
 // @flow
+import {Readable, Transform} from "stream";
 
 import SequentialIDGenerator from "../sequential-id-generator.js";
 
@@ -30,32 +31,44 @@ export type attrsT = {
 };
 
 export type sheetT = {|
+  insertBeforeMedia?: string,
   css: string,
   attrs: attrsT,
 |};
 
 type optionsT = {
   prefix?: string,
+  streamingMode?: boolean,
 };
 
 class StyletronServer implements StandardEngine {
   styleCache: MultiCache<{pseudo: string, block: string}>;
   keyframesCache: Cache<KeyframesObject>;
   fontFaceCache: Cache<FontFaceObject>;
-  styleRules: {[string]: string};
+  styleRules: {
+    [string]: {
+      insertBeforeMedia?: string,
+      rules: string,
+    },
+  };
   keyframesRules: string;
   fontFaceRules: string;
+  streamingMode: boolean;
 
   constructor(opts?: optionsT = {}) {
-    this.styleRules = {"": ""};
+    this.streamingMode = opts.streamingMode || false;
+    this.styleRules = {"": {rules: ""}};
     this.styleCache = new MultiCache(
       new SequentialIDGenerator(opts.prefix),
-      media => {
-        this.styleRules[media] = "";
+      (media, _cache, insertBeforeMedia) => {
+        this.styleRules[media] = {
+          ...(insertBeforeMedia && {insertBeforeMedia}),
+          rules: "",
+        };
       },
       (cache, id, value) => {
         const {pseudo, block} = value;
-        this.styleRules[cache.key] += styleBlockToRule(
+        this.styleRules[cache.key].rules += styleBlockToRule(
           atomicSelector(id, pseudo),
           block,
         );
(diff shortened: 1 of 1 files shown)


packages/styletron-engine-atomic/src/server/server.js
```diff
@@ -117,12 +130,74 @@ class StyletronServer implements StandardEngine {
...
+
+    const readableStream: Readable = htmlStream;
+    const CLOSING_TAG_R = /^\s*<\/[a-z]/i;
+
+    const instance = this;
+
+    const transformer = new Transform({
+      transform: function appendStyleChunks(chunk, /* encoding */ _, callback) {
+        // Get the chunk and retrieve the sheet's CSS as an HTML chunk,
+        // then reset its rules so we get only new ones for the next chunk
```
airhorns, line 175: Which piece actually does the reset for the next chunk?

airhorns:
@rtsao would love to know what you think of this approach!

packages/styletron-engine-atomic/src/server/server.js
AyoubElk: This is where that's handled:

https://github.com/AyoubElk/styletron/blob/01edeb3dd4031d963cd0c1b69520663e2fd377a9/packages/styletron-engine-atomic/src/server/server.js#L140-L147

airhorns:
or maybe @tajo might be able to say if this is a satisfactory direction? I think Ayoub is waiting on confirmation before writing tests and docs so it'd be awesome to get a "looks about right" or a "don't think we should do it this way" from you folks before investing more in this approach!

airhorns:
Also worth mentioning is that React is building out more and more support for streaming server side rendering, especially to support Suspense and the other stuff that Server Components might bring. See https://twitter.com/dan_abramov/status/1375275660994883588?s=20 for a reference to all the work they are putting in. I feel like this feature for Styletron would be great to land to have a strategy for working in the streaming future and for working better with those of us rendering very big pages that still desire a good TTFB!

packages/styletron-engine-atomic/src/server/server.js
```diff
@@ -176,17 +261,70 @@ function stringify(styleRules, sortedCacheKeys) {
...
+      const predecessor = document.querySelectorAll('[media="${sheet.insertBeforeMedia}"]')[0]
+      document.head.insertBefore(styleElement, predecessor);`);
+  } else {
+    content.push(`
+      document.head.append(styleElement);`);
+  }
+
+  content.push(`
+    }
+    styleElement.innerHTML = styleElement.innerHTML + " ${sheet.css}"
```
rtsao, line 321:
```suggestion
    styleElement.innerHTML = styleElement.innerHTML + " ${sheet.css}";
```

packages/styletron-engine-atomic/src/server/server.js
```diff
@@ -176,17 +261,70 @@ function stringify(styleRules, sortedCacheKeys) {
...
+      styleElement = document.createElement("style");`);
+
+  for (const attr in attrs) {
+    content.push(`
+      styleElement.setAttribute("${attr}", "${attrs[attr]}");`);
+  }
+
+  if (sheet.insertBeforeMedia) {
+    content.push(`
+      const predecessor = document.querySelectorAll('[media="${sheet.insertBeforeMedia}"]')[0]
```
rtsao, line 312: Probably best to avoid const/let. Also accessing the first element of `querySelectorAll` is equivalent to `querySelector`, right?

```suggestion
      var predecessor = document.querySelector('[media="${sheet.insertBeforeMedia}"]');
```

packages/styletron-engine-atomic/src/server/server.js
```diff
@@ -30,32 +31,44 @@ export type attrsT = {
...
-    this.styleRules = {"": ""};
+    this.streamingMode = opts.streamingMode || false;
+    this.styleRules = {"": {rules: ""}};
     this.styleCache = new MultiCache(
       new SequentialIDGenerator(opts.prefix),
-      media => {
-        this.styleRules[media] = "";
+      (media, _cache, insertBeforeMedia) => {
+        this.styleRules[media] = {
+          ...(insertBeforeMedia && {insertBeforeMedia}),
```
rtsao, line 65: I'm thinking it should be OK if the value of this property is undefined

```suggestion
          insertBeforeMedia,
```

packages/styletron-engine-atomic/src/server/server.js
```diff
@@ -176,17 +261,70 @@ function stringify(styleRules, sortedCacheKeys) {
...
+  <script>
+  (function () {
+  `);
+
+  const selector = attrs.media
+    ? `style.${className}[media='${attrs.media}']`
+    : `style.${className}:not([media])`;
+
+  content.push(`
+    let styleElement = document.head.querySelector("${selector}");
```
rtsao, line 301:
```suggestion
    var styleElement = document.head.querySelector("${selector}");
```

rtsao (commented):
This is pretty interesting, thanks for putting this together! A couple thoughts:

It's definitely a bit unfortunate JS is required for this. CSP is another factor that could make this approach tricky. There probably needs to be a way to control the attributes of the injected script tags for this reason (e.g. supply a nonce).

I'd definitely be interested in seeing how this mechanism of style injection would affect performance. I'm still wondering if it'd be better to simply disallow or issue warnings for broken edge cases (sibling selectors, etc.) with some kind of progressive rendering "strict mode". But it's certainly cool that this works for all styles.

packages/styletron-engine-atomic/src/server/server.js
AyoubElk: Yep, nice catch! I had fixed it locally afterwards but forgot to push that commit

AyoubElk:
> This is pretty interesting, thanks for putting this together!

Thanks!

> It's definitely a bit unfortunate JS is required for this. CSP is another factor that could make this approach tricky. There probably needs to be a way to control the attributes of the injected script tags for this reason (e.g. supply a nonce).

Understood, I'm thinking we could allow the user to provide a callback function that would be called for each generated script tag and used to do all sorts of things (including calculating and supplying a nonce).

I feel this would be generic enough to cover the CSP use case, and potentially other not so straight-forward use cases users might have.

What do you think?

> 
> I'd definitely be interested in seeing how this mechanism of style injection would affect performance.

How do you propose we test performance? I haven't seen any performance-related tests/benchmarks anywhere.

> I'm still wondering if it'd be better to simply disallow or issue warnings for broken edge cases (sibling selectors, etc.) with some kind of progressive rendering "strict mode". But it's certainly cool that this works for all styles.

I think issuing warnings or disallowing edge cases would be tricky to get right. I will think about that some more and see if I come up with a clean way to achieve it.

airhorns:
> I feel this would be generic enough to cover the CSP use case, and potentially other not so straight-forward use cases users might have.

I agree, I think if you made the function actually responsible for generating the script tag and then had a default that made a simple one that would give a lot of control. 

> I'd definitely be interested in seeing how this mechanism of style injection would affect performance.

Do you mean page load performance for the user experience or server side rendering performance for the backends doing SSR?

> I'm still wondering if it'd be better to simply disallow or issue warnings for broken edge cases (sibling selectors, etc.) with some kind of progressive rendering "strict mode". But it's certainly cool that this works for all styles.

To do this we'd have to parse and understand CSS selectors that styletron can find right, and then it'd still be hard to warn for CSS that is coming from some other spot on the page like a plain old `link` tag, right? That seems like a fair amount of complexity but I guess it'd only have to run in development.

AyoubElk:
I added tests and support for providing a custom script generation function.

For client-side hydration tests, any thoughts on the best way to test it? 
The hydration can only be done after the script tags code is executed and style tags have been generated

AyoubElk:
@rtsao It would be great to know what you think about this.

PR #406, opened 2021-07-07
Added speedy option when initialising monolithic client

liegeandlief:
I need to be able to disable the speedy insertion of styles using CSSOM when initialising the monolithic engine prior to passing it to the `styletron-react` provider. This is so that I can ensure that style tags get added to the DOM when prerendering my app.

This PR simply adds the `speedy` option to the monolithic engine options and passes it in when initialising a new `StyleSheet` which already has the `speedy` option.

Please see https://emotion.sh/docs/ssr#puppeteer for context as to why I would like this option.

### DIFF
--- a/packages/styletron-engine-monolithic/src/client/client.js
+++ b/packages/styletron-engine-monolithic/src/client/client.js
@@ -33,6 +33,7 @@ type optionsT = {
   container?: Element,
   prefix?: string,
   strict?: boolean,
+  speedy?: boolean,
 };
 
 type cacheT = {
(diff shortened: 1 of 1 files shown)


liegeandlief:
@tajo please can you take a look at this PR or tag somebody else to take a look? Thanks :)

liegeandlief:
@tajo Is it possible for somebody to take a look at this PR? It is only a small change which I don't think will take long to review and would be very helpful for me. 🤞 for a reply as this project seems to be very quiet lately, I'm hoping it is not dead!

tajo (approved):

tajo:
lgtm, only @rtsao can publish this

liegeandlief:
Thanks @tajo and @rtsao. Please can a new version of the `styletron-engine-monolithic` package now be published to NPM? Is there anything I need to do?

PR #413, opened 2021-09-30
Refactor DevProvider to function component

behnammodi:
All test passed


chasestarr:
Thank you for the contribution @behnammodi Let's keep the component as-is. There is no plan for react to remove class components

PR #418, opened 2022-02-11
Sequential id can now generate from a to ac

SukkaW:
Previously the sequential id generator prevent `ad` from being generated by starting with `ae` (347). After the PR, the sequential id can also generate from `a` to `ac` (skip `ad`).

See also https://github.com/styletron/styletron/pull/219

cc @tajo @rtsao

### DIFF
--- a/packages/styletron-engine-atomic/src/__tests__/test.node.js
+++ b/packages/styletron-engine-atomic/src/__tests__/test.node.js
@@ -6,6 +6,8 @@ import {validateNoMixedHand} from "../validate-no-mixed-hand.js";
 
 import sortMq from "../sort-css-media-queries.js";
 
+import SequentialIDGenerator from "../sequential-id-generator.js";
+
 test("validateNoMixedHand", t => {
   t.deepEqual(
     validateNoMixedHand({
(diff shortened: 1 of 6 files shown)


tajo:
Thanks but not sure if this really matters. 

1. The headache of getting all classNames renamed in a big codebase when using jest snapshots.
2. We primarily use the monolithic engine these days.

tajo:
> styletron-engine-monolithic is not battle-tested (uber.com & baseweb design still uses atomic engine)

We use it in 198 applications right now with no known issues so you can consider it battle-tested.

> styletron-engine-monolithic will double the size of emitted styles, which can not be underestimated

The output is bigger but with brotli compression it's not that significant (other resources like JS are usually a magnitude bigger).

> styletron-engine-monolithic has no typescript types support, which also suggests that the React community has very few experiences with it

It has the same API as atomic engine. You can copy&paste these types over.

SukkaW:
> The output is bigger but with brotli compression, it's not that significant (other resources like JS are usually a magnitude bigger).

The CSS declaration itself can be easily fit in brotli compression, but the randomized generated class name can not. Inline style is also critical for the first render/paint while "other resources like JS" can be deferred and progressively hydrate the entire application.

And some interesting details about Chrome: https://stackoverflow.com/questions/63869348/cumulative-layout-shift-with-bootstrap-4-grid. This StackOverflow question demonstrates how the large inline style can contribute to CLS (Cumulative Layout Shift).

SukkaW:
@tajo It appears that my PR has been left without a response for years. I presume that you and your team have denied merging the PR.

As I am no longer using styletron in any of my current projects, there is no reason for me to continue pushing this PR forward. Therefore, I am now closing the PR. If you or your team desire for the proposed change to be included in styletron, please don't hesitate to ping me and I am willing to help in resolving any merging conflicts.

PR #420, opened 2022-03-04
Refactor: use stylis as prefixer

SukkaW:
Recently, Robin Weser, the author of `inline-style-prefixer`, who is also the author of [Fela Atomic CSS-in-JS library](https://fela.js.org), has changed the Fela's CSS prefixer to [stylis](https://github.com/thysultan/stylis), an extremely lightweight CSS preprocessor that is also used by [emotion Monolithic CSS-in-JS library](https://emotion.sh).

By changing from `inline-style-prefixer` to `stylis`:

- The bundle size can be reduced, as stylis is very lightweight and fast (take a look at [how stylis' prefixer is implemented](https://github.com/thysultan/stylis/blob/master/src/Prefixer.js))
- stylis support a wider range of browsers without sacrificing the bundle size
- stylis' `prefix` function take in a string of CSS declaration and the length of the hyphenated property and return a string of prefixed CSS declaration string, which makes the code clean and fast (Previously, styletron has to make an adapter that creates a `StyleObject` for `inline-style-prefixer` from a string, and build a CSS declaration string from the returned prefixed `StyleObject` outputs)

----

~~Note: The PR is marked as a draft since stylis hasn't brought up the grid support (https://github.com/thysultan/stylis/pull/285). I will update the test case when the PR is merged and a new version of stylis is released.~~

### DIFF
--- a/packages/styletron-engine-atomic/src/inject-style-prefixed.js
+++ b/packages/styletron-engine-atomic/src/inject-style-prefixed.js
@@ -4,7 +4,8 @@ declare var __DEV__: boolean;
 
 import hyphenate from "./hyphenate-style-name.js";
 import {validateNoMixedHand} from "./validate-no-mixed-hand.js";
-import {prefix} from "inline-style-prefixer";
+
+import {prefix} from "stylis";
 
 import type {StyleObject} from "styletron-standard";
 
@@ -30,9 +31,9 @@ export default function injectStylePrefixed(
         validateValueType(originalVal, originalKey);
       }
 
-      const propValPair = `${hyphenate(
-        originalKey,
-      )}:${((originalVal: any): string)}`;
+      const hyphenatedProperty = hyphenate(originalKey);
+
+      const propValPair = `${hyphenatedProperty}:${((originalVal: any): string)}`;
       const key = `${pseudo}${propValPair}`;
       const cachedId = cache.cache[key];
       if (cachedId !== void 0) {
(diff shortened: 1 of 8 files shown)


packages/styletron-engine-atomic/src/inject-style-prefixed.js
```diff
@@ -41,27 +42,12 @@ export default function injectStylePrefixed(
...
-            }
-          }
-        }
-        block += propValPair; // ensure original prop/val is last (for hydration)
+        const cssDeclaration = `${hyphenatedProperty}:${((originalVal: any): string)};`;
+        const prefixed = prefix(cssDeclaration, hyphenatedProperty.length);
+        const block =
+          prefixed[prefixed.length - 1] === ";"
+            ? prefixed.slice(0, -1)
+            : prefixed;
```
SukkaW, line 50: `stylis` guaranteed that the original declaration will be at the end of the prefixed declaration.

SukkaW:
`stylis@4.1.0` is released with grid prefix support. The PR is now ready for review.

packages/styletron-engine-atomic/src/inject-style-prefixed.js
rtsao: Is this documented in stylis somewhere? It might be worth linking to the relevant part of stylis or at least mentioning it via a comment.

rtsao:
Thanks for putting this together, this seems like this would be a nice improvement!

 A couple questions:
- How much smaller does this make the styletron bundle?
- My understanding is stylis is super fast, I assume it is faster than inline-style-prefixer?
- How does browser support compare? You mention stylis has broader support, but I noticed `-moz-box` got removed from output. I assume this is totally deprecated?

SukkaW:
> * How much smaller does this make the styletron bundle?

So here are the size differences of `styletron-engine-atomic`:

- index.es.js (build from master branch): 26.219KiB
- index.es.js  (build from this PR): 25.492KiB
- browser.es5.es.js (build from master branch): 31.015KiB
- browser.es5.es.js (build from this PR): 30.260KiB

> * My understanding is stylis is super fast, I assume it is faster than inline-style-prefixer?

`stylis` is at least as fast as `inline-style-prefixer`, if not faster. It has a smaller size, and its prefixer is specifically optimized for both size and speed.

And IMHO size matters a lot for styletron. Unlike other Atomic CSS-in-JS solutions (such as Atlassian's [compiled css-in-js](https://github.com/atlassian-labs/compiled) or Facebook's StyleX) which generate CSS beforehand and most of the runtime will be stripped during the generation, styletron is Atomic CSS-in-JS runtime and the prefixer has to be bundled together.

> * How does browser support compare?

`stylis` is used by some very popular CSS-in-JS libraries like [emotion](https://emotion.sh) (which, here I quote, "supports all popular browsers, including Internet Explorer 11"), [linaria](https://github.com/callstack/linaria) and Vercel's [styled-jsx](https://github.com/vercel/styled-jsx).

And the support for prefixing CSS grid for IE 11 is added to `stylis` after https://github.com/thysultan/stylis/pull/285 & https://github.com/thysultan/stylis/pull/287 are merged and a new version is released.

> You mention stylis has broader support, but I noticed `-moz-box` got removed from output. I assume this is totally deprecated?

`-moz-box` only supports from Firefox 2 to Firefox 63 (see [data on caniuse](https://caniuse.com/mdn-css_properties_display_xul_box_values)). From Firefox 64 the `display: -moz-box` is completely ignored by default. On the other hand, `display: flex` is fully supported on Firefox since Firefox 28.

packages/styletron-engine-atomic/src/inject-style-prefixed.js
SukkaW: I don't believe it is necessary.

The previous code was originally for adapting `inline-style-prefixer`, which doesn't guarantee the order of prefixed, while stylis has the same behavior as `autoprefixer`, which is "correct".

SukkaW:
@rtsao @tajo It appears that my PR has been left without a response for years. I presume that you and your team have denied merging the PR.

As I am no longer using styletron in any of my current projects, there is no reason for me to continue pushing this PR forward. Therefore, I am now closing the PR. If you or your team desire for the proposed change to be included in styletron, please don't hesitate to ping me and I am willing to help in resolving any merging conflicts.

Issue #436, opened 2023-09-28
Add support for @container queries


MarkReeder:
It would be great to have @container query support since they are now supported across all major browsers:
https://developer.mozilla.org/en-US/docs/Web/CSS/@container#browser_compatibility

PR #437, opened 2023-09-29
Add @container query support

MarkReeder:
Addresses: https://github.com/styletron/styletron/issues/436

@container queries started landing in stable browsers about a year ago and have been fully supported across all major browsers for over 6 months now. I had a use case where they will be very useful and I wanted to use them with Styletron.

### DIFF
--- a/packages/styletron-engine-atomic/src/inject-style-prefixed.ts
+++ b/packages/styletron-engine-atomic/src/inject-style-prefixed.ts
@@ -77,13 +77,16 @@ export default function injectStylePrefixed(
             media,
             pseudo + originalKey,
           );
-      } else if (originalKey.substring(0, 6) === "@media") {
+      } else if (
+        originalKey.substring(0, 6) === "@media" ||
+        originalKey.substring(0, 10) === "@container"
+      ) {
         classString +=
           " " +
           injectStylePrefixed(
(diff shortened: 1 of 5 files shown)


rtsao (approved):

packages/styletron-engine-atomic/src/inject-style-prefixed.ts
```diff
@@ -77,13 +77,16 @@ export default function injectStylePrefixed(
...
+        originalKey.substring(0, 6) === "@media" ||
+        originalKey.substring(0, 10) === "@container"
+      ) {
         classString +=
           " " +
           injectStylePrefixed(
             styleCache,
             originalVal as StyleObject,
-            originalKey.substr(7),
+            originalKey,
```
rtsao, line 89: To be honest, I don't remember the original reasoning for not storing the at-prefix. But looking at this now, this seems preferable since it avoids having to get the substring and also avoids the later concatenation upon injection

MarkReeder: Great! I considered just dropping the substring matching entirely and just checking if the first character was `@`, but that seemed beyond the scope of this change, but might be worth considering at some point.

